PR #249, opened 2013-08-10
allow custom boolean attributes

amitport:
Should allow custom boolean attributes (the are legal HTML).

Note that currently, in order to set a custom boolean attribute you need to write something like: elm[0].attribs['name'] = null. It would be nice to implement http://api.jquery.com/prop/ at some point.

### DIFF
--- a/lib/render.js
+++ b/lib/render.js
@@ -8,11 +8,6 @@ var decode = utils.decode;
 var encode = utils.encode;
 
 /*
-  Boolean Attributes
-*/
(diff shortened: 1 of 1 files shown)


triglian:
Hello, this is a real world use case that we have faced:

With Web Components you can have Custom Elements with custom properties that can be set through attributes. Example code:

```html
<my-element id="myEl"></my-element>
```

```js
$("#myEl").attr("showconfidence", false);
$.html("#myEl")
// prints <my-element id="myEl" showconfidence="false"></my-element>
```

In the above example, when the resulting element is rendered clientside, the value of the `showconfidence` attribute is a String.

It would be great if  `showconfidence` was treated as a boolean attribute from cheerio. thanks!

Issue #315, opened 2013-12-03
CSS parsing does not handle url(http://foo); correctly


Meekohi:
If your element has `background-image: url(http://foo.com/image.jpg);` cheerio will parse it as `{'background-image':'http'}` due to splitting on the : character.

PR #316, opened 2013-12-03
Ignore colons inside of url(.*) when parsing css

Meekohi:
Parsing by css is totally wrong, but this will at least not break on `url(http://whatever);` anymore! Closes issue #315.


chiqui3d:
is still not working for mi case https://github.com/cheeriojs/cheerio/issues/1223

PR #515, opened 2014-06-10
.attr setting boolean attributes

0xBADC0FFEE:
Fixed setting boolean attributes


fb55:
This has been open for a long while; closing this as I don't think we want to handle boolean attributes this way. They should not actually have boolean values.

PR #588, opened 2014-11-02
Ensure text nodes expose the DOM level 1 API

jugglinmike:
Since enabling the `withDomLvl1` parsing option, nodes cannot be created
with an object literal. Create new text nodes using the `evaluate`
function to ensure they expose the correct attributes.


fb55:
Closing in favour of #1161

PR #619, opened 2015-01-01
createWriteStream method

LinusU:
Exposes `.createWriteStream(options)` which takes the same options as `.load(..., options)` and returns a writeable stream.

A `finish` event is emitted when the dom is ready.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -83,4 +83,8 @@ exports.update = function(arr, parent) {
   return parent;
 };
 
+exports.createWriteStream = function (options, cb) {
+  return htmlparser.createDomStream(cb, options);
+};
+
 // module.exports = $.extend(exports);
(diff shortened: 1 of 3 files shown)


fb55:
I would prefer a `stream(cb, opts)` method (shorter name, extended signature), with both arguments being optional. `finish` and `error` are quite nice though.

Mickael-van-der-Beek:
The Travis CI's report sas that there were no conflicts and that the tests passed so I think this can be be merged to the master branch ?

jugglinmike:
@Mickael-van-der-Beek Our approval criteria are a little more subjective. See [Felix's comment above](https://github.com/cheeriojs/cheerio/pull/619#issuecomment-68502084) and [the pull request he has made since then](https://github.com/cheeriojs/cheerio/pull/637).

PR #620, opened 2015-01-01
load from url

LinusU:
*Depends on #619, pull it first.*

Adds support for `.load(url, cb)` which streams an html document from the web into the parser. The callback receives `(err, $)`.

Sample program:
```js
var cheerio = require('./');

cheerio.load('http://example.com/', function (err, $) {
  if (err) { throw err; }

  console.log($('h1').text());
});
```

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -83,4 +83,8 @@ exports.update = function(arr, parent) {
   return parent;
 };
 
+exports.createWriteStream = function (options, cb) {
+  return htmlparser.createDomStream(cb, options);
+};
+
 // module.exports = $.extend(exports);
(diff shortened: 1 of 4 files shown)


jugglinmike:
While the underlying functionality proposed by gh-619 seems in line with Cheerio's responsibilities, this feels like an over-extension. It introduces another library (which depends on "once" which depends on "wrappy") and simply wraps it.

Web page scraping is definitely a common use case for Cheerio, but it's not the only use case. Increasing this library's responsibilities like this has a few drawbacks:

- The `load` method is harder to understand/document
- The maintainers of Cheerio have to manage "simple-get" as a dependency and inherit some responsibility for its bugs (including its nested dependencies)
- The `load` method takes a slight performance hit

I guess we could mitigate two of those problems by introducing a dedicated static method for this, but I still think this violates the single responsibility principle. I appreciate the effort, though; I'd just much rather share this code as an example usage.

davidchambers:
> I'd just much rather share this code as an example usage.

Agreed. :)

PR #626, opened 2015-01-11
Add coveralls badge, remove link to old report

fb55:
It wasn't obvious to me which version of cheerio the old report was about. This way, we have a concrete, automatically updated coverage report.


davidchambers:
LGTM

Issue #621, opened 2015-01-03
Replace .text()


alitavector:
It may sound strange to replace api ```.text()``` as ```.t```. I guess there a cleaner way to write less code and for my clarity sake. How do I use it?

``` $$.find("choice").text() ```
into
``` $$.f("choice").t ```

**Comment by davidchambers** [comment]
jQuery compatibility is important, so we won't change the name of this method.

**Comment by jugglinmike** [comment]
David's right, but [you can now do this yourself](https://github.com/cheeriojs/cheerio#plugins)! Here's how:

```js
$$.prototype.f = $$.prototype.find;

Object.defineProperty($$.prototype, 't', {
  get: function() {
    return this.text();
  }
});
```

Issue #589, opened 2014-11-03
Cheerio plugins


Yomguithereal:
Good evening,
Sorry if this question has already been addressed somewhere else but I did not find the relevant issue or thread.

Is there a generally admitted method for extending cheerio's methods? Is overloading `cheerio.protoype` the correct way to perform this?

``` js
// Registering a plugin
var cheerio = require('cheerio');
cheerio.prototype.scrape = function() {
  // custom logic...
}
```

PR #629, opened 2015-01-15
Isolate prototypes of functions created via `load`

jugglinmike:
I wanted to shore up the project's recommendation for extending Cheerio and back it with documentation and tests. This will allow us to be more confident in recommending this pattern in the future. Related issues: gh-621 & gh-589

Commit message:

> Ensure that each Cheerio function created with the `load` static method
> has a dedicated object in its prototype chain. This allows for
> conflict-free extension of the generated prototype.
>
> Additionally, expose the `fn` prototype alias as defined by jQuery and
> used by many jQuery plugins.


davidchambers:
:deciduous_tree: Beautiful patch.

fb55:
What @davidchambers said. :deciduous_tree:

jugglinmike:
Thanks, guys! :evergreen_tree:

Issue #129, opened 2012-12-01
$("tr:eq(2)") won't work but $("tr").eq(2) does 


tekny:
The problem is `:eq`, `:gt`, `:lt` selectors are not working. They just return empty result.

PR #630, opened 2015-01-15
Add .gt() and .lt()

wataori:
#129

We've wanted to use .gt() and .lt() like jQuery and I added these features.

I use the ` Array.slice() ` function to add these .gt() and .lt() functions.
these functions may be different to the concept of the cheerio if it is ` O(n) `, but I don't know :-(
Please tell us if someone know about this.

### DIFF
--- a/lib/api/traversing.js
+++ b/lib/api/traversing.js
@@ -353,6 +353,14 @@ exports.eq = function(i) {
   return this[i] ? this._make(this[i]) : this._make([]);
 };
 
+exports.gt = function(i) {
+  return this.length > 1 ? this._make(this.slice(i + 1)) : this;
+};
+
+exports.lt = function(i) {
+  return this.length > 1 ? this._make(this.slice(0, i)) : this;
+};
+
 // Retrieve the DOM elements matched by the jQuery object.
 exports.get = function(i) {
   if (i == null) {
--- a/Readme.md
+++ b/Readme.md
@@ -493,6 +493,21 @@ $('li').eq(-1).text()
 //=> Pear
 ```
 
+#### .gt( i )
+Select all elements at an index greater than index within the matched set.
+
+```js
+$('li').gt(1)
+//=> Pear
+```
+
+#### .lt( i )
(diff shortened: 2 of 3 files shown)


lib/api/traversing.js
```diff
@@ -353,6 +353,14 @@ exports.eq = function(i) {
   return this[i] ? this._make(this[i]) : this._make([]);
 };
 
+exports.gt = function(i) {
+  return this.length > 1 ? this._make(this.slice(i + 1)) : this;
```
fb55: I'm not sure about the behavior for `length <= 1`. Shouldn't an empty collection be returned?

jugglinmike:
These methods are not implemented in jQuery, so including them here would only allow people to write less portable code. I know you are making up for the lack of the associated pseudo selectors (an existing portability concern), but the best response to that would be to implement those in the CSSselect library.

That may be a little more challenging, but the good news is that we now have [explicit instructions for extending Cheerio](https://github.com/cheeriojs/cheerio#plugins). You can follow those to implement any custom API you would like for your application, including `lt` and `gt`.

lib/api/traversing.js
wataori: Uhhh...Exactly. Thanks.

wataori:
> These methods are not implemented in jQuery, 

Do you mean that these are implemented in Sizzle? I'm a little confused. But OK. I'd like to define and try these methods at prototype object in my own apps. I'll push commits if these get succesful.

jugglinmike:
@wataori Correct. For instance, there is [a `:gt` selector](http://api.jquery.com/gt-selector/), but there is no `gt` method. As noted in the relevant jQuery documentation pages, if you'd like to use a method, `slice` offers equivalent behavior. And good news: [Cheerio implements the `slice` method](https://github.com/cheeriojs/cheerio#slice-start-end-) :)

wataori:
@jugglinmike 
>  Cheerio implements the slice method :)

Great!!

PR #235, opened 2013-07-05
Implement .serializeArray()

jlep:
jQuery compatible .serializeArray() implementation (http://api.jquery.com/serializeArray/) + test cases. Doesn't handle multiple selections yet.

Issue #69, opened 2012-06-30
.serialize()?


rgarcia:
http://api.jquery.com/serialize/

this would be nice to have when finding hidden inputs etc.

fb55:
I guess this is long fixed :)

PR #631, opened 2015-01-19
Fixed up `serializeArray()` and added multiple support

twolfson:
Fixes #235 

PR #235 has been open for a while and stagnated. This PR revives adding `serializeArray()` support and establishes a foundation for other form methods like `serialize` (#69). In this PR:

- Fixed up merge conflicts from #235 
- Fixed up broken implementation due to time difference
- Added support for multiple selected items (a la jQuery) as requested in #235 
- Various style changes to be more consistent with the rest of the repo


fb55:
@twolfson Thanks for picking this up from @jlep!

Some notes:

- Please be more generous with white space.
- It's nice to point to sources, but that can be part of the commit message. While a single source as part of a comment is still fine, you've overdone it a bit.
- Please add comments instead regarding the function of code. Keep in mind that the code needs to be maintained indefinitely once this pull request is merged, and you might not be around to explain it.
- Please remove dead code (instead of commenting it out).

lib/api/serialize.js
```diff
@@ -0,0 +1,32 @@
...
+      var type = $elem.attr('type');
+      return $elem.attr('name') && !$elem.is(':disabled') &&
+        // known because of `.find`
+        /* rsubmittable.test(this.nodeName) && */ !rsubmitterTypes.test(type) &&
+        ($elem.attr('checked') || !rcheckableType.test(type));
+    }).map(function(i, elem) {
+      var $elem = Cheerio(elem);
+      var name = $elem.attr('name');
+      var val = $elem.val();
+      return val == null ?
```
fb55: This would be much more readable in an `if / else` block.

lib/api/serialize.js
```diff
@@ -0,0 +1,32 @@
...
+        null :
+        Array.isArray(val) ?
+          _.map(val, function(val) {
+            return {name: name, value: val.replace( rCRLF, '\r\n' )};
```
fb55: What's the rationale behind encoding all line breaks this way?

lib/api/serialize.js
```diff
@@ -0,0 +1,32 @@
...
+    // rsubmittable = /^(?:input|select|textarea|keygen)/i;
+exports.serializeArray = function() {
+  var Cheerio = this.constructor;
+  return this.find('input,select,textarea,keygen') // Taken from `rsubmittable`
+    .filter(function() {
+      // https://github.com/jquery/jquery/blob/2.1.3/src/serialize.js#L91-L94
+      // TODO: Come back to disabled
+      var $elem = Cheerio(this);
+      var type = $elem.attr('type');
+      return $elem.attr('name') && !$elem.is(':disabled') &&
```
fb55: I wanted to add the `:matches` selector for some time now, which would allow the inclusion of this in the selector:

```css
:matches(input, select, textarea, keygen)
[name]
:not(:disabled)
:not(:checkbox, :radio)
:matches([checked], :not(

```

fb55: It might be a good idea to pre-compile the selector in this case, though, so let's forget that for now.

twolfson:
Have you considered adding `jscs` to enforce whitespace? I have found it very useful on all of my projects (usually no back and forth about stylistic issues).

http://jscs.info/rules.html

I prefer to leave sources inline since the references to the original code can fade away as more touches occur and `git blame` erodes. However, I will respect your request.

Most of this code is copy/paste from jQuery but I will do my best to document it.

As for the inline dead code, that was also inline from jQuery but I will make a comment note instead.

lib/api/serialize.js
twolfson: That was directly from jQuery. I will break it up.

lib/api/serialize.js
twolfson: This is from jQuery. I am guessing on different OS' and different browsers, the line breaks inside of textareas aren't consistent. This standardizes them to `\r\n`. Should we keep it? I will definitely add a comment.

lib/api/serialize.js
twolfson: Alright, leaving this chunk alone.

lib/api/serialize.js
jugglinmike: I don't know if you're set up to run jQuery's unit tests (it's a little bit of a hassle), but if you are, you might try removing the replacement and seeing which tests break.

If you can add a unit test that exercises the code and that is logical, then keep it. Otherwise we shouldn't be maintaining code we don't need/understand.

lib/cheerio.js
```diff
@@ -13,7 +13,8 @@ var api = [
   require('./api/attributes'),
   require('./api/traversing'),
   require('./api/manipulation'),
-  require('./api/css')
+  require('./api/css'),
+  require('./api/serialize')
```
jugglinmike: Could you rename the new source file to "forms.js"? This matches jQuery's documented category scheme and will make a logical home for the `serializeArray` function in the future.

twolfson: Sure thing. This was originally named `api.serialize.js` by @jlep so I tried to keep it as consistent as possible.

lib/api/serialize.js
twolfson: I will take a look at the tests files first. It does look a bit painful.

twolfson: Fwiw, it looks like someone was working on setting up Vagrant which should make everything less painful.

https://github.com/jquery/jquery/pull/1410

twolfson: I have found the corresponding tests:

https://github.com/jquery/jquery/blob/2.1.3/test/unit/serialize.js#L130-L136

https://github.com/jquery/jquery/blob/2.1.3/test/index.html#L152-L153

When I inspect the value of a `<textarea>` in Firefox on Linux Mint, it only has a `\n`, no `\r\n`:

```js
encodeURIComponent($($0).val());
// "hello%0A%20%20world"
```

I will add a test for this to prevent regressions.

twolfson:
Alright, I have renamed the file so unfortunately most of the inline comments have gone away. I have addressed all of them.

twolfson:
I have re-pushed the branch because Travis CI choked on `node@0.11` for some reason.

twolfson:
Ah, never mind it isn't me. It looks like `master` is having a build error.

fb55:
At this point I would be okay with merging this.

The code would be much simpler when using the `:matches` selector, but that's something I'll need to implement.

twolfson:
Sweet, I see that `master` has been patched. I am going to wiggle my PR build.

twolfson:
Alright, the PR has been wiggled and Travis CI is now green.

jugglinmike:
I think this is almost there; it's just missing one piece of functionality that I can see. From the jQuery docs:

> This method can act on a jQuery object that has selected individual form controls, such as `<input>`, `<textarea>`, and `<select>`.

That means before we land this, the patch should be updated to pass a test like:

```js
+       it('() :', function() {
+         expect($('#nested input').serializeArray()).to.eql([
+            { name: 'fruit', value: 'Apple' },
+            { name: 'vegetable', value: 'Carrot' }
+          ]);
+       });

```

@twolfson It's been a while since anybody gave you feedback on this. If you'd moved on to bigger and better things, just let me know and I'll be happy to finish this up for ya

twolfson:
Sure thing. I have a bit of time to kill before heading out tonight. I will take a shot at those new tests and comment when this PR is updated.

jugglinmike:
That's some stick-to-itiveness for you! Thanks :)

twolfson:
Alright, I have added a new test and fix. I also saw that the tests are very unclear about what results should be. I have refactored them to be much more about asserting deep equals of arrays/objects to make the return value more obvious.

lib/api/forms.js
```diff
@@ -0,0 +1,55 @@
...
+  return this.map(function() {
+      var elem = this;
+      var $elem = Cheerio(elem);
+      if (elem.name === 'form') {
+        return $elem.find(submittableSelector).toArray();
+      } else {
+        return $elem.filter(submittableSelector).toArray();
+      }
+    })
+    .filter(function() {
```
jugglinmike: Could you move this method call to the previous line for visual consistency in the chain?

twolfson: Whoops, sure thing

twolfson: Updated

jugglinmike:
Rebased and merged to `master` via gh-669. Thanks for sticking with us, Todd!

twolfson:
Woot, no problem. Glad to see it got landed. There was also the follow-up PR #632 which I will update in a bit to have a much cleaner git history.

PR #632, opened 2015-01-20
Added `.serialize()` support. Fixes #69

### LINKED ISSUE #69: .serialize()?

twolfson:
Now that #631 is implemented, we can add `.serialize()` support. Unfortunately, this comes with adding `$.param()`. `$.param` has a second half that `.serialize()` doesn't leverage (see its `else` statement). I am not sure whether we want to add this/test it, or drop it and make `$.param` internal only for now.

In this PR:

- Added `.serialize()` support http://api.jquery.com/serialize/
- Added tests for `.serialize()`

### DIFF
--- a/lib/api/forms.js
+++ b/lib/api/forms.js
@@ -2,9 +2,23 @@
 // https://github.com/jquery/jquery/blob/2.1.3/src/serialize.js
 var _ = require('lodash'),
     submittableSelector = 'input,select,textarea,keygen',
+    r20 = /%20/g,
     rCRLF = /\r?\n/g,
     rcheckableType = /^(?:checkbox|radio)$/i,
     rsubmitterTypes = /^(?:submit|button|image|reset|file)$/i;
+
+exports.serialize = function() {
+  // Convert form elements into name/value objects
+  var arr = this.serializeArray();
+
+  // Serialize each element into a key/value string
+  var retArr = _.map(arr, function(data) {
+    return encodeURIComponent(data.name) + '=' + encodeURIComponent(data.value);
+  });
+
+  // Return the resulting serialization
+  return retArr.join('&').replace(r20, '+');
+};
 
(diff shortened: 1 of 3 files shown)


twolfson:
I have caught up this PR with the changes from landing #631 to `master`. It looks like I was wreckless when adding tests (i.e. didn't test all of jQuery's functionality) and was focused on using `serialize` to get a query string. I will revisit this later.

twolfson:
I have chosen to take the shorter path and stripped out everything from `jQuery.param` that is irrelevant to `.serialize()`. The result is a much smaller PR. There is 1 line I haven't been able to figure out how to test which is:

```js
var value = data.value == null ? '' : data.value;
```

I cannot seem to find a case when `data.value == null`. `.serializeArray()` returns no data for `val == null`.

https://github.com/cheeriojs/cheerio/blob/ccd9369153e9c580782ae422eff59d6caed16f1d/lib/api/forms.js#L38

fb55:
@twolfson I can't think of a case where data.value is `null` either, just drop the line. The rest LGTM.

twolfson:
I have updated the PR to ditch the `null` check.

mh166:
@twolfson, @fb55: is there any news on this PR? It would be very useful to me in my current project. :+1:

twolfson:
There's a merge conflict and I no longer have write access to the source repo. I'm going to close and reopen this PR.

@mh166 You can point to a `git` URL for dependencies in `npm`:

https://docs.npmjs.com/files/package.json#urls-as-dependencies

```js
"cheerio": "git://github.com/underdogio/cheerio#dcd66809"
```

twolfson:
I have opened the replacement PR #827

mh166:
Thanks! It's greatly appreciated!

PR #637, opened 2015-01-25
`.stream` method

fb55:
Extended version of #619. Still needs more tests, other feedback would be appreciated, though.

### DIFF
--- a/lib/static.js
+++ b/lib/static.js
@@ -26,7 +26,7 @@ exports.load = function(content, options) {
   // Ensure that selections created by the "loaded" `initialize` function are
   // true Cheerio instances.
   initialize.prototype = Cheerio.prototype;
-  // Keep a reference to the top-level scope so we can chain methods that implicitly 
+  // Keep a reference to the top-level scope so we can chain methods that implicitly
   // resolve selectors; e.g. $("<span>").(".bar"), which otherwise loses ._root
   initialize.prototype._originalRoot = root;
 
@@ -39,6 +39,23 @@ exports.load = function(content, options) {
   initialize._options = options;
 
   return initialize;
+};
+
+/**
+* $.createWriteStream()
+*/
+
+exports.createWriteStream = function(options) {
+
+  var stream = parse.createWriteStream(options, function done(err, dom) {
+    if (err) {
+      stream.emit('error', err);
+    } else {
+      stream.emit('finish', exports.load(dom));
+    }
+  });
+
+  return stream;
 };
 
 /**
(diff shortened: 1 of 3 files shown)


lib/static.js
```diff
@@ -50,6 +50,36 @@ exports.load = function(content, options) {
...
+      stream.emit('finish', loaded);
+
+      if(typeof cb === 'function'){
+        cb(null, loaded);
+      }
+    }
+  }, options);
+
+  if(typeof cb === 'function'){
+    stream.on('error', cb);
```
jugglinmike: We could do the same for the "finish" event

jugglinmike:
Cool! I pushed a commit to make sure that parsing options are honored by the generated constructor (plus a few extra tests). Hope that wasn't too forward of me.

I'd prefer to keep this as simple as possible and let the user attach event handlers if they need a callback. At the same time, I don't write that much streams code, so if you think it's idiomatic to expose a callback, then I'm fine with it.

lib/static.js
fb55: The first argument of the callback is supposed to be an error object, the callback would need to be wrapped anyway. Adding it to the stream callback seems to be the best idea.

jugglinmike: Sounds good

fb55:
I was thinking about it from a usability standpoint. IMHO a nice way to process a webpage would be

```js
request("http://google.com")
  .on("error", handleError)
  .pipe(cheerio.stream(function(err, $){
    if(err) return handleError();
    …
  }));
```

Of course, this could still be written as

```js
request("http://google.com")
  .on("error", handleError)
  .pipe(cheerio.stream())
    .on("error", handleError)
    .on("finish", function($){
      …
    });
```
…which is quite nice, too, even though I prefer the first solution. Providing both options allows everybody to decide on their own. I'm okay with removing the callback though.

> Hope that wasn't too forward of me.

I appreciate it!

pedro93:
Any update on this?

mscdex:
+100 I would love to see cheerio support a streaming parser. That's really the only solution for really large documents (maybe not so much HTML, but XML for example).

fb55:
The stream API would only support parsing (and potentially allow some early querying), the document would still be stored in memory. A module like cornet or trumpet might be a better fit for that.




On Fri, Nov 18, 2016 at 3:31 AM +0100, "Brian White" <notifications@github.com> wrote:












+100 I would love to see cheerio support a streaming parser. That's really the only solution for really large documents (maybe not so much HTML, but XML for example).

waynevanson:
I would absolutely love to see full streaming support for an API like cheerio. I'm currently researching how to write something like this for use in a streaming fashion.

I'm still shopping around on ideas and implementations and will comment if I end up doing anything.

fb55:
Closing this PR as the approach is outdated. Still interested to revisit eventually.

PR #639, opened 2015-01-25
Allow failures in Node.js v0.11

jugglinmike:
I think allowing failures here is the correct policy in principle, but we now have a practical reason to do so. From https://github.com/tmpvar/jsdom/issues/984 (Dec 20, 2014):

> I am not interested in taking on the support burden of working with unstable Node versions. Wait for iojs.

Commit message:

> Node.js version 0.11 is an unstable platform, meaning it may introduce
> breaking changes at any time. It is useful for the project maintainers
> to be aware of such failures (because they represent potential problems
> with the next stable release of Node.js), they should not block ongoing
> work.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,3 +5,5 @@ node_js:
 script: make travis-test
 matrix:
   fast_finish: true
+  allow_failures:
+    - node_js: 0.11


davidchambers:
LGTM

jugglinmike:
@fb55 This reverts part of [a change you made](https://github.com/cheeriojs/cheerio/commit/20578ca39af28b5a22e66bbc2143f89002021fe7), so I'd like to get your feedback before landing it.

jugglinmike:
Also thanks David :)

fb55:
As long as failures are due to upstream changes, this seems to be the way to go.

FWIW, we might want to set `sudo = false`, to allow execution inside of a container.

PR #642, opened 2015-01-28
Add a newline for 'br' element while converting to text()

harish2704:
Add a newline for 'br' element while converting to text()

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -113,7 +113,9 @@ var readData = function(el, name) {
       } else if (value === String(Number(value))) {
         value = Number(value);
       } else if (rbrace.test(value)) {
-        value = JSON.parse(value);
+          try {
+              value = JSON.parse(value);
+          } catch(e){ }
       }
 
       el.data[jsName] = value;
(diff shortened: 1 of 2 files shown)


fb55:
While I understand the intent, that's not how jQuery behaves, which is our reference. I'm +1 for the fix in the first commit, though.

jugglinmike:
I agree with @fb55 . While the `<br>` tag is typically used to insert a line break, that is not necessarily it's effect on the document. For example:

```css
br { display: none; }
```

...but that first commit is a nice idea. @harish2704 do you think you could open a new pull request with that change and a unit test?

fisch0920:
Any chance this functionality could be enabled optionally with a flag passed to `.text()`?

I think it's what most people expect, and it's really hard to mimic the expected behavior without this option built into cheerio.

fb55:
I'm closing this in favor of #676.

@fisch0920 Here is a replacement function:

```js
function textWithNewlines($elems){
  return $elems.toArray().map(function serialize(e){
    if(e.type === "text") return e.data;
    if(e.type === "tag" || e.type === "script" || e.type === "style"){
      return (e.name === "br" ? "\n" : "") + e.children.map(serialize).join("");
    }
    return "";
  }).join("");
}
```

PR #651, opened 2015-02-04
Updates Readme.md: .not(function (index, elem))

### LINKED ISSUE #650: Readme example for .not( function(index, elem) ) should call .not()
The example in `Readme.md` for the `.not( function(index, elem) )` function calls `.filter( (function(index, elem) )` when it should actually call `.not( (function(index, elem) )`.

```js
$('li').filter(function(i, el) {
  // this === el
  return $(this).attr('class') === 'orange';
}).length;
//=> 2
```

This actually returns `1`.

The listed example should instead be:

```js
$('li').not(function(i, el) {
  // this === el
  return $(this).attr('class') === 'orange';
}).length;
//=> 2
```

This correctly returns `2`.

plward11:
Updates the Readme.md example given for the .not(function (index, elem))
function to correctly call .not(function (index, elem)).

Fixes #650.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -441,7 +441,7 @@ $('li').not('.apple').length;
 Function:
 
 ```js
-$('li').filter(function(i, el) {
+$('li').not(function(i, el) {
   // this === el
   return $(this).attr('class') === 'orange';
 }).length;


davidchambers:
Thanks!

plward11:
No problem, glad to help :)

PR #653, opened 2015-02-11
Update lodash@3.1.0

SamyPesse:
This PR updates lodash to the version  `3.1.0`. The version `2.4.1` was crashing in web worker.

PR #647, opened 2015-02-02
Recognize options in $.xml()

fb55:
Also moved shared logic with `.html()` to a function.

fixes #645

PR #655, opened 2015-02-13
Change lodash dependencies to ^3.2.0

SamyPesse:
I know I made a PR for this a few days ago (#653). But I think `~3.1.0` is too strict considering what cheerio is using from lodash (core methods): `_.each`, `_.forEach`, `_.map`, `_.defaults` and `_.merge`; these methods are not planned to change in the `3.x.x`  or anytime soon.

Being less strict is better for people using it with something like browserify (reduce bundle size by avoiding having 2 versions of lodash).

For example, from `3.1.0` to `3.2.0`, jdalton fixed some "important" issues in non-core methods (chaining): https://github.com/lodash/lodash/issues/955 and https://github.com/lodash/lodash/issues/958.

Another, maybe better and even less strict, solution can be: `3.x.x || 2.4.x`

Let me know what you think.

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,7 +25,7 @@
     "entities": "~1.1.1",
     "htmlparser2": "~3.8.1",
     "dom-serializer": "~0.0.0",
-    "lodash": "~3.1.0"
+    "lodash": "^3.2.0"
   },
   "devDependencies": {
     "benchmark": "~1.0.0",


SamyPesse:
@fb55 any news on when the next version of cheerio will be released ?

SamyPesse:
?

fb55:
There are some prs I'd like to see merged before publishing a new version (eg. #631 and #647). I'm hesitant to go ahead with them without an okay from @MatthewMueller, @jugglinmike or @davidchambers.

jugglinmike:
@fb55 sorry about those lingering ones--I'll take a look this weekend.

Issue #657, opened 2015-02-18
incorrect .map example on website


geobrando:
So I'm new to cheerio/jquery and javascript in general.  I've been going through the docs at http://cheeriojs.github.io/cheerio/ and couldn't figure out why I was unable to get the the following .map example to work.

    $('li').map(function(i, el) {
      // this === el
      return $(this).attr('class');
    }).join(', ');
    //=> apple, orange, pear

Then I found the same example at [readme.md](/cheeriojs/cheerio/blob/master/Readme.md) which shows the example as ```.get().join(', );``` and works fine.  I was going to submit a pull request and fix this but realized that the site doesn't appear to be maintained anymore.

@MatthewMueller Are there plans to put up a site using the readme?  I noticed a cheeriojs.github.io repo was created last  year but is currently empty.

PR #658, opened 2015-02-18
add get method to .map example

geobrando:
As mentioned in #657, not sure if this site is being maintained anymore.  But this example matches the example on the readme and works as intended.


fb55:
I'm not sure what the state of the site is, this patch is an improvement nevertheless :)

PR #661, opened 2015-02-21
Add Table of Contents to README

samosfator:
How about a Table of Contents for easier navigation through API docs?


samosfator:
Maybe later. I realised that anchors not working for code snippets.

fb55:
The readme is quite large, a TOC would definitely be a good idea.

Issue #490, opened 2014-05-23
Support of $.prop()


nodecode:
I have just replaced jQuery in my Node.js application by Cheerio, and could see a huge acceleration. So far, everything works well, except for the function <code>$.prop()</code>, which is not available in Cheerio. 

It would be good if Cheerio would support <code>$.prop()</code>.

http://api.jquery.com/prop/

fb55:
Fixed in #668

PR #668, opened 2015-03-01
Prop method PR updated

rwaldin:
Updated version of PR #516 rebased against latest master.  I think I've address feedback to the original PR, added a bit more to the implementation of .prop('style'), plus tests.  For issue #490.

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -101,10 +101,6 @@ exports.prop = function (name, value) {
 
         property.length = i;
 
-        property.item = function(index) {
-          return this[index];
-        };
-
         break;
       case 'tagName':
       case 'nodeName':
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -73,6 +73,69 @@ exports.attr = function(name, value) {
   return getAttr(this[0], name);
 };
 
+var getProp = function (el, name) {
+  return el.hasOwnProperty(name)
+      ? el[name]
+      : rboolean.test(name)
+          ? getAttr(el, name) !== undefined
+          : getAttr(el, name);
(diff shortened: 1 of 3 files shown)


lib/api/attributes.js
```diff
@@ -73,6 +73,75 @@ exports.attr = function(name, value) {
...
+};
+
+exports.prop = function (name, value) {
+  var i = 0,
+      property;
+
+  if (typeof name === 'string' && value === undefined) {
+
+    switch (name) {
+      case 'style':
```
jugglinmike: Cheerio's responsibilities stop at the jQuery API; mimicking DOM API's in this library is a slipperly slope. For instance, why `item()` but not `removeProperty()`? A more complete implementation might find it's home in [the DOMHandler library](https://github.com/fb55/domhandler) which would trickle down to Cheerio. It feels like an over-extension to insert a partial implementation here.

rwaldin: Sounds reasonable.  Updated PR to remove item() method and tests.

lib/api/attributes.js
```diff
@@ -73,6 +73,71 @@ exports.attr = function(name, value) {
...
+        property = this.css();
+
+        _.each(property, function (v, p) {
```
rwaldin: If you prefer, I can remove the setting of property names and length on the property object...  Or to go even further, remove the special handling of getting 'style' and 'tagName|nodeName' altogether.  Any of those options would be reasonable in order to get a prop() method.

ApXaHgheJI:
When the prop will work ?

fb55:
@rwaldin Sorry for the delay – and thanks!

PR #670, opened 2015-03-01
Use a selector to filter form elements

fb55:
Follow-up for #631.


jugglinmike:
Very nice! I've squashed these commits together and merged them to `master` at 541815a2bd59cb583ea1de1213b04f23e1639cc3 .

PR #671, opened 2015-03-01
Added support for options without a `value` attribute. Fixes #633

### LINKED ISSUE #633: Select value doesn't fallback to text content when `value` is undefined
In browsers, `el.value` for an `<option>` will fallback from `getAttribute('value')` to the inner content of a node. Since jQuery executes in the browser, it gets this for free when it asks for `el.value`. However, Cheerio currently does not support this behavior. As a result:

```html
<select><option selected>Hello</option></select>
```

`$('select').val()` will return `[undefined]` via Cheerio and `['Hello']` via jQuery.

twolfson:
As noted in #633, we are not properly treating `option's` without a `value` attribute. This PR adds back functionality that emulates the DOM and thus jQuery behavior. In this PR:

- Added tests for value-less options (e.g. normal, with HTML entities, with HTML elements)
    - Correct expectations can be found at http://codepen.io/twolfson/pen/WbymMK
- Added patches for `attr` and `val`

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -1,4 +1,5 @@
 var _ = require('lodash'),
+  $ = require('../static'),
   utils = require('../utils'),
   isTag = utils.isTag,
   domEach = utils.domEach,
@@ -37,6 +38,11 @@ var getAttr = function(elem, name) {
   if (hasOwn.call(elem.attribs, name)) {
     // Get the (decoded) attribute
     return rboolean.test(name) ? name : elem.attribs[name];
+  }
+
+  // Mimic the DOM and return text content as value for `option's`
+  if (elem.name === 'option' && name === 'value') {
+    return $.text(elem.children);
   }
 };
 
@@ -198,7 +204,7 @@ exports.val = function(value) {
       if (this.attr().hasOwnProperty('multiple')) {
         returnValue = [];
         domEach(option, function(i, el) {
-          returnValue.push(el.attribs.value);
+          returnValue.push(getAttr(el, 'value'));
         });
       }
       return returnValue;
--- a/test/api/attributes.js
+++ b/test/api/attributes.js
@@ -286,6 +286,18 @@ describe('$(...)', function() {
       var val = $('select#one').val();
       expect(val).to.equal('option_selected');
     });
+    it('.val(): on select with no value should get text', function() {
+      var val = $('select#one-valueless').val();
+      expect(val).to.equal('Option selected');
+    });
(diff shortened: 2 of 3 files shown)


test/api/attributes.js
```diff
@@ -308,7 +320,11 @@ describe('$(...)', function() {
     });
     it('.val(): on multiple select should get an array of values', function() {
       var val = $('select#multi').val();
-      expect(val).to.have.length(2);
+      expect(val).to.have.eql(['2', '3']);
```
fb55: `.to.eql` would be more appropriate :)

fb55:
Besides the unnecessary `have` inside two tests: LGTM.

test/api/attributes.js
twolfson: But I can English good =(

Heh, nice catch. I will fix it this weekend.

twolfson:
I have updated the PR with the requested changes.

lib/api/attributes.js
```diff
@@ -1,10 +1,12 @@
 var _ = require('lodash'),
+  static = require('../static'),
```
davidchambers: `static` is [reserved][1] as a future keyword. Perhaps we should use `static_` instead.


[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#Future_reserved_keywords

jugglinmike: Better yet, we could avoid defining the variable altogether:

```diff
-  static = require('../static'),
   // ...
-  text = static.text,
+  text = require('./static').text,
```

Besides side-stepping the naming problem, this makes the file a little more maintainable--it's clear that *only* the exported `text` method is used here (as opposed to "anything in the `static` module").

twolfson: I used this style for consistency with the file as is. I suggest that if we make that change then all other `utils` should have a similar format:

```js
   isTag = require('../utils').isTag,
   domEach = require('../utils').domEach,
   hasOwn = Object.prototype.hasOwnProperty,
   camelCase = require('../utils').camelCase,
   cssCase = require('../utils').cssCase,
```

twolfson: It looks like we are using `$` for static in other files. I am going to switch to that for consistency for now:

https://github.com/cheeriojs/cheerio/blob/0.18.0/lib/api/manipulation.js#L3

twolfson:
I have updated the PR with the requested changes.

fb55:
@twolfson Sorry for the delay – and thanks!

PR #672, opened 2015-03-02
Added default value for checkboxes/radios

twolfson:
While working on #632 to find some edge cases jQuery is covering, I found a regression with checkboxes and radio inputs. In the browser, it will default to `on` as the value when no value attribute has been specified.

http://codepen.io/twolfson/pen/bNKJBZ

This PR fixes that by patching in on `.val()`. In this PR:

- Added regression test for valueless radios and checkboxes
- Added fix for valueless radios and checkboxes

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -44,6 +44,13 @@ var getAttr = function(elem, name) {
   if (elem.name === 'option' && name === 'value') {
     return $.text(elem.children);
   }
+
+  // Mimic DOM with default value for radios/checkboxes
+  if (elem.name === 'input' &&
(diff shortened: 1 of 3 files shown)


fb55:
LGTM. This is in-line with the HTML spec (both value attributes are set to [default/on](http://www.w3.org/html/wg/drafts/html/master/#dom-input-value-default-on)) & the overhead shouldn't be too bad.

fb55:
@twolfson Sorry for the delay. This needs a rebase, should be good to merge then.

twolfson:
This PR has been updated and the merged conflicts have been resolved.

fb55:
Awesome, thanks :)

PR #674, opened 2015-03-02
Add missing optional selector doc for {prev,next}{All,Until}

astorije:
Looking at [the code](https://github.com/cheeriojs/cheerio/blob/d41d8954f747c074e621d884e346f2f420df59de/lib/api/traversing.js#L143) or [tests](https://github.com/cheeriojs/cheerio/blob/d41d8954f747c074e621d884e346f2f420df59de/test/api/traversing.js#L212), I can see that these are there and tested, so why not advertise them! :-)
It also matches [the jQuery docs](http://api.jquery.com/nextAll/) now.

Thanks!

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -309,16 +309,18 @@ $('.apple').next().hasClass('orange')
 //=> true
 ```
 
-#### .nextAll()
-Gets all the following siblings of the first selected element.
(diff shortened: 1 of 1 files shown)


fb55:
Thanks!

PR #676, opened 2015-03-03
Handle malformed JSON in data attributes

fb55:
The fix was part of #642, I've added a test case.

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -113,7 +113,9 @@ var readData = function(el, name) {
       } else if (value === String(Number(value))) {
         value = Number(value);
       } else if (rbrace.test(value)) {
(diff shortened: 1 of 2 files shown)


lib/api/attributes.js
```diff
@@ -113,7 +113,9 @@ var readData = function(el, name) {
       } else if (value === String(Number(value))) {
         value = Number(value);
       } else if (rbrace.test(value)) {
-        value = JSON.parse(value);
+          try {
```
davidchambers: :triangular_ruler: Alignment looks off.

fb55: Good call, fixed.

davidchambers:
:deciduous_tree:

PR #678, opened 2015-03-03
fix #677: attr({foo: null}) removes attribute foo, like attr('foo', null)

### LINKED ISSUE #667: Declaring JavaScript string literals that contain <script> tags within a <script> tag are not parsed correctly
Example:

Contents of file print_script_tag_text.js:
```js
$ = cheerio.load("<script>var foo = '<script>closing script tag disappears</script>';</script>");
$.root().find('script').each(function(context, element) {
    console.log($(this).text());
});
```
$> node print_script_tag_text.js
```js
var foo = '<script>closing script tag disappears
```

### LINKED ISSUE #677: attr({foo: null}) should remove the attribute foo as does attr('foo', null)
Calling attr({foo: null}) should remove the attribute foo, but instead sets the attribute foo="null"

rwaldin:
also added unit tests for both attr(name, null) and attr({name: null}) cases...

Fixes #677

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -61,8 +61,8 @@ exports.attr = function(name, value) {
       if (!isTag(el)) return;
 
       if (typeof name === 'object') {
-        _.each(name, function(name, key) {
-          el.attribs[key] = name+'';
+        _.each(name, function(value, name) {
+          setAttr(el, name, value);
         });
       } else {
         setAttr(el, name, value);
--- a/test/api/attributes.js
+++ b/test/api/attributes.js
@@ -111,6 +111,22 @@ describe('$(...)', function() {
       $apple.removeAttr('data-autofocus');
       expect($apple.attr('data-autofocus')).to.be(undefined);
     });
+
+    it('(key, value) : should remove attributes when called with null value', function() {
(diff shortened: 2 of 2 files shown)


fb55:
LGTM.

MichielDeMey:
+1

jugglinmike:
@arthurvr [The jQuery API documentaion on `attr`](https://api.jquery.com/attr/#attr2) does not include information on unsetting attributes. Given that [jQuery has unit tests for that behavior](https://github.com/jquery/jquery/blob/f5aa89af7029ae6b9203c2d3e551a8554a0b4b89/test/unit/attributes.js#L508-L517), I believe this is a mistaken omission. Could you confirm?

arthurvr:
@jugglinmike Hey! No worries, [we're planning to document that behaviour](https://github.com/jquery/api.jquery.com/issues/523). But thanks for asking!

jugglinmike:
Good news! [We've recently documented this feature in jQuery](https://github.com/jquery/api.jquery.com/pull/853), so it's now safe to implement in Cheerio. Merged to `master` at commit b27bed6c679cc8beea913ab6b7e6269ca6445b11. Thanks for the patch (and for your patience), @rwaldin!

Issue #703, opened 2015-04-21
.html(number) and .text(number) break things


Rycochet:
Not sure exactly what's going on - but looks like when they're set they're not being converted into a string, so when you try to change them later it's not behaving properly.

(Pseudo code as this was found in a larger project)

$("span").html(123).html(123) silently fills it with an empty string.

$("span").text(123).text(123) causes an exception -

```
Warning: Object 312 has no method 'replace' Use --force to continue.
TypeError: Object 312 has no method 'replace'
    at Object.encodeXML (...\node_modules\cheerio\node_modules\entities\lib\encode.js:58:6)
    at renderText (...\node_modules\cheerio\node_modules\dom-serializer\index.js:166:21)
    at module.exports (...\node_modules\cheerio\node_modules\dom-serializer\index.js:121:17)
    at renderTag (...\node_modules\cheerio\node_modules\dom-serializer\index.js:146:14)
    at module.exports (...\node_modules\cheerio\node_modules\dom-serializer\index.js:113:17)
    at renderTag (...\node_modules\cheerio\node_modules\dom-serializer\index.js:146:14)
    at module.exports (...\node_modules\cheerio\node_modules\dom-serializer\index.js:113:17)
    at renderTag (...\node_modules\cheerio\node_modules\dom-serializer\index.js:146:14)
    at module.exports ...\node_modules\cheerio\node_modules\dom-serializer
\index.js:113:17)
    at renderTag (...\node_modules\cheerio\node_modules\dom-serializer\inde
x.js:146:14)
```

Both `$("span").html("" + 123).html("" + 123)` and `$("span").text("" + 123).text("" + 123)` work as expected.

Maybe coerce the type at some point?

fb55:
PRs welcome :)

fb55:
Fixed in #704

PR #704, opened 2015-04-22
Fix for passing non-string values to .html or .text

Rycochet:
Fix for #703


jugglinmike:
Thanks for the patch! Would you mind adding tests for this change?

fb55:
@Rycochet Sorry for the delay – and thanks!

PR #707, opened 2015-05-14
Solved "Cannot read property 'parent' of undefined"

dokson:
Solving this:

parse.js: at line 55
    var oldParent = node.parent || node.root,
                        ^
TypeError: Cannot read property 'parent' of undefined

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -2,6 +2,7 @@
   Module Dependencies
 */
 var htmlparser = require('htmlparser2');
+var util = require('util');
 
 /*
   Parser
@@ -50,33 +51,42 @@ exports.update = function(arr, parent) {
   // Update neighbors
   for (var i = 0; i < arr.length; i++) {
     var node = arr[i];
+    if ( !util.isNullOrUndefined(node)) {
+        // Cleanly remove existing nodes from their previous structures.
+        var oldParent = null;
+        var oldSiblings = null;
+        if (!util.isNullOrUndefined(node.parent)) {
+            oldParent = node.parent;
+            oldSiblings = oldParent && oldParent.children;
+        } else if (!util.isNullOrUndefined(node.root)) {
+            oldParent = node.root;
+            oldSiblings = oldParent && oldParent.children;
 
-    // Cleanly remove existing nodes from their previous structures.
-    var oldParent = node.parent || node.root,
-        oldSiblings = oldParent && oldParent.children;
-    if (oldSiblings && oldSiblings !== arr) {
-      oldSiblings.splice(oldSiblings.indexOf(node), 1);
-      if (node.prev) {
-        node.prev.next = node.next;
-      }
-      if (node.next) {
-        node.next.prev = node.prev;
-      }
-    }
+        }
+        if (oldSiblings && oldSiblings !== arr) {
+            oldSiblings.splice(oldSiblings.indexOf(node), 1);
+            if (node.prev) {
+                node.prev.next = node.next;
+            }
+            if (node.next) {
+                node.next.prev = node.prev;
+            }
+        }
 
-    if (parent) {
-      node.prev = arr[i - 1] || null;
-      node.next = arr[i + 1] || null;
-    } else {
-      node.prev = node.next = null;
-    }
+        if (parent) {
+            node.prev = arr[i - 1] || null;
+            node.next = arr[i + 1] || null;
+        } else {
+            node.prev = node.next = null;
+        }
 
-    if (parent && parent.type === 'root') {
-      node.root = parent;
-      node.parent = null;
-    } else {
-      node.root = null;
-      node.parent = parent;
+        if (parent && parent.type === 'root') {
+            node.root = parent;
+            node.parent = null;
+        } else {
+            node.root = null;
+            node.parent = parent;
+        }
     }
   }
 


davidchambers:
Please add tests. :)

dokson:
I'm sorry, but I'm noob to NodeJs.
Just solving this error got using https://www.npmjs.com/package/kickstarter-crawler 
(https://github.com/ghostsnstuff/kickstarter-crawler)

jugglinmike:
Can you share the code that triggered the error? We need to understand the bug that you are trying to fix, and then we'll want to be sure it doesn't come back.

inDream:
@jugglinmike I've faced this bug with these conditions:
`cheerio.load(null)` or `cheerio.load(undefined)`
But this would not trigger the bug: `cheerio.load(false)`

jugglinmike:
@inDream I'm not sure those examples qualify as "bugs" because they describe undocumented usages of the `load` method. It might be nice if Cheerio threw a more descriptive error message, but generally speaking, it doesn't seem useful or intuitive to "load" values like `null` or `undefined`.

I might be wrong, though--can you give some context as to why you would like to use `load` in that way?

inDream:
@jugglinmike If load data from remote endpoint, network problem would cause this bug. I have no idea why this would happen as I haven't log the response content. Maybe there's other case would cause this bug but this is only one of the example and I haven't test this whether would fix the bug.

jugglinmike:
@inDream If the data source is unreliable, you might consider doing some explicit error checking in your application logic *prior* to invoking `cheerio.load`. That way you can recover or exit gracefully at the point of failure (and not bother attempting to parse/manipulate invalid data). Does that sound like a better approach to you?

inDream:
@jugglinmike This is a better approach but my concern is the error message is not clear.

jugglinmike:
@inDream Fair enough. The solution to your problem would take a completely different form than the patch provided here, though. Because it concerns an undocumented usage of the API, the fix is low-priority, but I would welcome a patch.

chubaka:
Hi,
I am getting this error:
  var oldParent = node.parent || node.root,
                        ^

TypeError: Cannot read property 'parent' of undefined
Can you help?

davidchambers:
Could you provide a [small, self-contained example][1], @chubaka?


[1]: http://sscce.org/

lfarquhar:
// I have the same issue. Here is an example.
var request = require('request');
var cheerio = require('cheerio');

var url = 'http:/www.msn.com';
request(url, function(err, resp, body){
    $ = cheerio.load(body);
});

davidbayo10:
This code is working for me.

``` javascript
const request = require('request');
const iconv = require('iconv-lite');

function getDOM(url) {
  return new Promise(function (resolve, reject) {
    request({ uri: url, encoding: null }, function (err, res, html) {
      if (err) {
        reject(err);
      } else {
        html = iconv.decode(html, 'ISO-8859-1');
        resolve(cheerio.load(html));
      }
    });
  });
}
```

manishsingh10895:
I am trying to load a local html file using **fs** and i was getting the same error, then I tried @davidbayo10 solution but still there's an error **cannot read property length of undefined** 
code for more info
```js 
    fs.readFile(p, (err, data)=> {
        HandleDataStructureResponse(err,{},data);
    });

 function HandleDataStructureResponse(err, res, body)
    {
        if(err) return console.log(err);
        let html = iconv.decode(body, 'ISO-8859-1');
        let $ = cheerio.load(body);
        var counter = 0;

        return [];
    }
```

Error log

```sh
<h1>Cannot read property 'length' of undefined</h1>
        <h2></h2>
        <pre>TypeError: Cannot read property 'length' of undefined
    at SBCSDecoder.write (/home/hp/Documents/Node/Geeky/node_modules/iconv-lite/encodings/sbcs-codec.js:61:32)
    at Object.decode (/home/hp/Documents/Node/Geeky/node_modules/iconv-lite/lib/index.js:38:23)
    at HandleDataStructureResponse (/home/hp/Documents/Node/Geeky/services/geekyParse.js:27:26)
    at Object.GetAllDataStructures (/home/hp/Documents/Node/Geeky/services/geekyParse.js:22:5)
    at /home/hp/Documents/Node/Geeky/routes/geeky.js:9:25
    at Layer.handle [as handle_request] (/home/hp/Documents/Node/Geeky/node_modules/express/lib/router/layer.js:95:5)
    at next (/home/hp/Documents/Node/Geeky/node_modules/express/lib/router/route.js:131:13)
    at Route.dispatch (/home/hp/Documents/Node/Geeky/node_modules/express/lib/router/route.js:112:3)
    at Layer.handle [as handle_request] (/home/hp/Documents/Node/Geeky/node_modules/express/lib/router/layer.js:95:5)
    at /home/hp/Documents/Node/Geeky/node_modules/express/lib/router/index.js:277:22</pre>
```

fb55:
Closing this as it is still missing a test case. If someone is able to provide one I am happy to reopen it.

softgrand:
TypeError: Cannot read property 'parent' of undefined
    at Function.exports.update (C:\Users\reza\Desktop\getomidname asasname\node_modules\cheerio\lib\parse.js:55:25)
    at module.exports (C:\Users\reza\Desktop\getomidname asasname\node_modules\cheerio\lib\parse.js:17:11)
    at Function.exports.load (C:\Users\reza\Desktop\getomidname asasname\node_modules\cheerio\lib\static.js:22:14)
    at Request._callback (C:\Users\reza\Desktop\getomidname asasname\type1asasomid.js:94:29)
    at self.callback (C:\Users\reza\Desktop\getomidname asasname\node_modules\request\request.js:186:22)
    at Request.emit (events.js:107:17)
    at ClientRequest.<anonymous> (C:\Users\reza\Desktop\getomidname asasname\node_modules\request\request.js:785:18)
    at ClientRequest.g (events.js:199:16)
    at ClientRequest.emit (events.js:104:17)
    at Socket.emitTimeout (_http_client.js:543:10)

softgrand:
help me

fb55:
Could you provide the HTML and the code that is causing that behavior?

PR #714, opened 2015-05-25
Update Readme.md

joepie91:
`.siblings` and `.children` have a function signature like `.parents`, but the `selector` argument was not listed as being optional in the former cases. This PR fixes that, and documents the function signature like that of `.parents`.

I'll probably submit some more documentation patches later, if I run across more issues.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -366,7 +366,7 @@ $('li').slice(1, 2).length
 //=> 1
 ```
 
-#### .siblings( selector )
+#### .siblings([selector])
 Gets the first selected element's siblings, excluding itself.
 
 ```js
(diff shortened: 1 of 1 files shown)


joepie91:
I should note that I'm not entirely clear on the correct documentation for `.closest`:

1. The function signature suggests `selector` being required, unlike `.parents`.
2. The function does not throw an error when the `selector` is left undefined, like `.parents`.
3. ... but it also doesn't return anything meaningful - just an empty array.

What's the intended behaviour/documentation for `.closest`, and would it need a similar change?

fb55:
Good catch, thanks!

As you said, `closest` doesn't find anything if no selector is provided (probably in-line with jQuery), but that's not really a use-case for the method.

Issue #541, opened 2014-07-20
Feature request: Implement $.wrap()


impinball:
This is actually a rather useful method, and would probably be even better server-side. It would simplify a lot of things server-side. It has been part of jQuery since 1.0, with the callback equivalent since 1.4.

htmljosh:
> 
> 
> A couple of my utility functions for cheerio - feel free to make them into proper methods ;-)
> 
> ```js
> function wrap(selector, wrapper) {
>     return $(selector).each(function() {
>         $(this).before(wrapper).prev().append(this);
>     });
> }
> 
> function unWrap(selector) {
>     $(selector).each(function() {
>         var $this = $(this);
>         $(this).after($this.contents()).remove();
>     });
> }
> 
> function wrapAll(selector, wrapper) {
>     $(selector).first().before(wrapper).prev().append(selector);
> }
> ```

Very helpful, thank you

PR #558, opened 2014-09-08
Wrap

dandlezzz:
Hey all,
I think I have made some progress on .wrap(), however, I cannot get it to work with strings, only cheerio nodes. This PR contains the failing test commented out. I saw other open issues regarding string input on .replaceWith();

Any tips?

Thanks,

PR #715, opened 2015-05-28
Implement `$.fn.wrap`

jugglinmike:
@dandlezzz started on this implementation in gh-558. I offered to finish up for him, got distracted by a more fundamental patch that it necessitated (see gh-574), then forgot about it. I've just now brought it back up to speed.

I've squashed Daniel's original work into  single commit and applied my own changes on top in order to clearly document this collaboration.

Supersedes gh-558
Resolves gh-541

### DIFF
--- a/lib/api/manipulation.js
+++ b/lib/api/manipulation.js
@@ -113,10 +113,10 @@ exports.wrap = function(wrapper) {
   var self = this,
   elemNumber = self.length;
 
-  _.times(elemNumber, function(i) {  
+  _.times(elemNumber, function(i) {
     if ( i > 0 ) wrapper = wrapper.clone();
-    
-    var el = self[i], 
+
+    var el = self[i],
         parent = el.parent || el.root,
         siblings = parent.children,
         dom = self._makeDomArray(typeof wrapper === 'function' ? wrapper.call(el, i, el) : wrapper),
@@ -126,7 +126,7 @@ exports.wrap = function(wrapper) {
 
     updateDOM(dom, null);
     index = siblings.indexOf(el);
-    
+
     var oldEl = uniqueSplice(siblings, index, 1, dom, parent);
     wrapper.prepend(oldEl);
     el.parent = el.prev = el.next = el.root = null;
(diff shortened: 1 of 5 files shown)


lib/api/manipulation.js
```diff
@@ -109,6 +110,40 @@ exports.prepend = _insert(function(dom, children, parent) {
...
+    }
+
+    dom = this._makeDomArray(wrapper, i < lastIdx).slice(0, 1);
+    index = siblings.indexOf(el);
+
+    updateDOM([el], dom[0]);
+    // The previous operation removed the current element from the `siblings`
+    // array, so the `dom` array can be inserted without removing any
+    // additional elements.
+    uniqueSplice(siblings, index, 0, dom, parent);
```
fb55: IMHO it would be simpler to reference just the element instead of `dom` and explicitly create an array for `uniqueSplice`.

fb55:
LGTM, except one tidbit.

lib/api/manipulation.js
jugglinmike: Hm, I don't follow. This is how I understand the operation graphically:

    dom:      [ d1, ..., dn ]

    parent:   p
    siblings:  \ [ s1, ..., el, ..., sn ]

to:

    parent:   p
    siblings:  \ [ s1, ..., d1, ..., dn, ..., sn ]
                             \ [ el ]

...which, to me, means I can't use `el` when splicing into `siblings`.

Can you share the code for the simplification that you had in mind?

fb55: `dom` contains only a single element (after the `.slice(0, 1)` in its initialization in L134). My proposal was to replace the slice with an access of the first element (`[0]`), which removes the access in L137, but requires a new array to be created in L141. As this is a stylistic choice, I'm fine with merging as-is. It tells a lot that this is the only thing I could find, great PR :)

jugglinmike: I was game to try this, but one of the `wrap` tests specifies a function as the only argument and does not return a value. The simplification you suggested causes an error to be thrown in this case because it assumes `dom` is an array with length >= 1. The expected behavior here is undefined, but since jQuery doesn't throw (and since this is a code style change more than anything), I think it makes sense to leave it as-is.

fb55: I discarded that case as _"Why would anyone ever do that?"_, guess I was wrong ^^

Thanks for looking into it. As no one else is weighting in, I'm merging this.

jugglinmike: Thanks for the review!

PR #718, opened 2015-06-09
interpret null in objects passed to attr() consistently

ssp:
… that is by removing the attribute in question rather than setting
it to the string "null".

Also add a test for this and avoid overwriting variables in the
implementation.

This should solve #677.

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -61,8 +61,8 @@ exports.attr = function(name, value) {
       if (!isTag(el)) return;
 
       if (typeof name === 'object') {
-        _.each(name, function(name, key) {
-          el.attribs[key] = name+'';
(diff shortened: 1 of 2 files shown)


fb55:
How does this differ from #678? (…which should have been merged by now.)

ssp:
@fb55 no functional difference, I guess (I just hadn't discovered that pull request)

#678 still overwrites the variable names which I find unnecessarily confusing (when »name« is an object, we overwrite »name« to be a key from that object?)

jugglinmike:
I'm not a big fan of shadowing either, but besides having been submitted first, gh-678 has an additional unit test and doesn't duplicate the logic for handling `null` values (note that the internal `setAttr` function already covers this concern). I've landed that one instead. Thanks anyway!

Issue #641, opened 2015-01-28
Implementation of appendTo and prependTo


YetiCGN:
While cheerio features some methods of DOM manipulation compatible with the jQuery API, it is missing [appendTo()](http://api.jquery.com/appendto/) and [prependTo()](http://api.jquery.com/prependto/), which are basically just flipped versions of append() and prepend().

I'd like to see this in cheerio, if it's not too much of a problem since I'm trying to test code which would need much refactoring to work without those methods.

fb55:
Pull requests welcome :)

tommedema:
+1 This would allow me to write

    $parent = $('<' + fragment.tag + '></' + fragment.tag + '>').attr(fragment.attribs) .appendTo($parent);

Instead of

       $newParent = $('<' + fragment.tag + '></' + fragment.tag + '>').attr(fragment.attribs);
        $parent.append($newParent);
        $parent = $newParent;

digihaven:
@jugglinmike is the pull request good?

fb55:
Fixed with #797

PR #726, opened 2015-06-24
Added appendTo and prependTo with tests see #641

digihaven:


lib/api/manipulation.js
```diff
@@ -102,6 +102,26 @@ var uniqueSplice = function(array, spliceIdx, spliceCount, newElems, parent) {
   return array.splice.apply(array, spliceArgs);
 };
 
+exports.appendTo = function(target) {
+	if (typeof target === 'string') {
```
jugglinmike: This project uses two-space soft tabs. Could you change the code style here to match?

test/api/manipulation.js
```diff
@@ -496,6 +496,24 @@ describe('$(...)', function() {
     });
   });
 
+  describe('.appendTo', function() {
+
+    it('(Node) : should add element as last child', function() {
```
jugglinmike: The test description is incorrect. It describes a Node-accepting signature, but the test invokes `appendTo` with a Cheerio instance (which is elsewhere described as `"$(...)"`)

test/api/manipulation.js
```diff
@@ -496,6 +496,24 @@ describe('$(...)', function() {
...
+      $('<li class="plum">Plum</li>').appendTo($fruits);
+      expect($fruits.children().eq(3).hasClass('plum')).to.be.ok();
+    });
+
+  });
+
+  describe('.prependTo', function() {
+
+    it('(html) : should add element as first child', function() {
```
jugglinmike: The test description is incorrect. It describes an HTML-accepting signature, but the test invokes `prependTo` with a Cheerio instance (which is elsewhere described as `"$(...)"`)

jugglinmike:
Thanks for the patch, @digihaven! Could you add some more tests? Some important ones that come to mind:

- Invoking with a string of HTML
- Invoking with a "node" instance
- Invoking with a selector
- Invoking with an Array

fb55:
Fixed with #797

PR #732, opened 2015-07-04
Reinstate `$.fn.toArray`

jugglinmike:
jQuery defines the `$.fn.toArray` method, and it is *not* deprecated, so
Cheerio should document and maintain it for the foreseeable future.


fb55:
This should have been merged long ago.

PR #734, opened 2015-07-18
Add a new FAQ field to highlight beginner.

jackzzjack:


fb55:
Might be a starting point for a future FAQ, although there could be some more content & actual questions.

Issue #733, opened 2015-07-08
Setting data on multiple elements doesn't work


magcius:
Simple reproducer:

```
var $ = Cheerio.load('<div></div> <div></div>');
var $div = $('div');
$div.data('foo', 'whatever');
```

gives me

```
/home/jstpierre/Random/node_modules/cheerio/lib/api/attributes.js:79
    el.data[name] = value;
                  ^
TypeError: Cannot set property 'foo' of undefined
    at setData (/home/jstpierre/Random/node_modules/cheerio/lib/api/attributes.js:79:19)
    at null.<anonymous> (/home/jstpierre/Random/node_modules/cheerio/lib/api/attributes.js:145:7)
    at exports.domEach (/home/jstpierre/Random/node_modules/cheerio/lib/utils.js:52:24)
    at exports.data (/home/jstpierre/Random/node_modules/cheerio/lib/api/attributes.js:144:5)
    at Object.<anonymous> (/home/jstpierre/Random/chrjs/test.js:6:6)
    at Module._compile (module.js:456:26)
    at Object.Module._extensions..js (module.js:474:10)
    at Module.load (module.js:356:32)
    at Function.Module._load (module.js:312:12)
    at Function.Module.runMain (module.js:497:10)
```

This seems to be because the `data` dictionary is only initialized for the first element if it's missing.

https://github.com/cheeriojs/cheerio/blob/master/lib/api/attributes.js#L128

PR #738, opened 2015-08-01
Enable setting data on all elements in selection

jugglinmike:
According to the description of `$.fn.data` (below), the method should
set the data for *all* elements in the selection. Ensure this does not
generate an error in those cases where any elements in the selection do
not yet define a `data` property.

> Store arbitrary data associated with the matched elements or return
> the value at the named data store for the first element in the set of
> matched elements.

Source: https://api.jquery.com/data/

This resolves gh-733


fb55:
Nice, simple patch.

PR #740, opened 2015-08-11
update the links to DOMHandler and Parser-options documentation and clarify the names of the modules

paulovieira:
The links for the htmlparser2 options are broken. This commit has the correct links.

Also, it makes clear the names of the modules. there is aN "htmlparser" module, but cheerio uses "htmlparser2" (this can be confusing the newcomers).


fb55:
Thanks!

PR #754, opened 2015-09-22
Fixed .html() to behave the same as in the documentation.

Animadei:


fb55:
Nice find, although we mirror jQuery's behavior and the current one is expected. I fixed the example in the readme instead -- I hope that's okay for you?

Animadei:
http://api.jquery.com/html/ shows the behavior you described. The documentation update is accepted. Thank you.

Issue #767, opened 2015-10-17
npmignore coverage report


nknapp:
I am currently trying to reduce the total installation size (and time) of my package `bootprint-swagger`. In order to do this, I inspect dependencies and look for unneeded files.

It seems to me, that the coverage report is included in the published version of cheerio. I would like to add the 'coverage/` directory to the `.npmignore`-file.

PR #739, opened 2015-08-03
Fix npm files filter

TrySound:
For smaller package

PR #768, opened 2015-10-17
Include coverage report in npmignore file (#767)

nknapp:


JaKXz:
:+1:

JaKXz:
but I think #739 is better.

nknapp:
I agree, #739 is better

PR #769, opened 2015-10-19
fixed require issue

daemonchen:
fixed require issue

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,4 +8,4 @@ exports = module.exports = require('./lib/cheerio');
   Export the version
 */
 
-exports.version = require('package.json').version;
+exports.version = require('./package.json').version;
--- a/index.js
+++ b/index.js
@@ -8,4 +8,4 @@ exports = module.exports = require('./lib/cheerio');
   Export the version
 */
 
-exports.version = require('./package').version;
+exports.version = require('package.json').version;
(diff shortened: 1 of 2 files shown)


daemonchen:
fixed some dependencies bug

fb55:
Which platform requires this?

jugglinmike:
We'll need more information before we can consider this for inclusion in `master`. Since we haven't heard back in over a month, I'm going to close this. We can re-open if the information comes to light!

ItsCosmo:
I am not the author of this PR, but please consider re-examining the "fixed require issue". I can explain why this fix is helpful. In my webpack configuration, I am using the "resolve" configuration as follows:
```
resolve {
  root: "./src"
  extension: ['', 'js', 'jsx', 'scss']
}
```
This is quite useful because is makes imports/requires simpler in the source code.

With the original line in cheerio/index.js:

`exports.version = require('./package').version`

The package.json file is not found, giving the error:

`Error: Cannot find module './package' from 'index.js'`

I believe that's because the empty extension is causing webpack to look in the "./src" folder for this file. Changing the line to:

`exports.version = require('./package.json').version`

The file is no longer matched to the resolve.extension, and so it loads properly. I have not been able to find a workaround for this issue. In the interest of making the integration of cheerio as seamless and trouble free as possible in a variety of environments and configurations, this change is warranted. After all, require("./package") is just a shorthand for require("./package.json")

luanmuniz:
I think the extension addition can go in. It look like this is causing problems and its a really simple change. I cant see any down-side.

The only problem with this PR is the changes in the `package.json` file. The owner of the PR add a few dependences for no reason.

PR #773, opened 2015-10-23
Test against node v0.12 --> v4.2

JaKXz:


fb55:
Thanks, definitely a welcome patch :)

This requires a jsdom update as well, even though AFAIK it is only required for benchmarking. It should probably be updated and moved to `optionalDependencies`.

`0.10` and `0.11` (including the `allow_failures` entry) can be removed, and `latest` should be added.

@JaKXz Could you update your patch to include these changes?

JaKXz:
@fb55 done :)

fb55:
Allowing failures for *unstable* is a good idea. Also, it should be
possible to use containers by passing *sudo: false* in the *.travis.yml*.

JaKXz:
@fb55 I think we're good now - would you like me to squash these commits? otherwise have fun and thanks for a great library :)

fb55:
@JaKXz Squashing the commits would be great, thanks a lot already :)

JaKXz:
@fb55 all done :)

fb55:
Awesome, thanks a lot!

Issue #728, opened 2015-06-29
css-select upgrade to 1.1.0 causing tests to fail


blondyn:
Hi, the change of css-select dependency is causing quite a few tests to fail. ( at least on my 2 osx systems) 29d38e32889417390ee682b4169d6f9eee5f9d14

jugglinmike:
Thanks for the report! [Our CI builds passed on that commit](https://travis-ci.org/cheeriojs/cheerio/builds/66485511), but [we're not currently testing in Node.js v0.12](https://github.com/cheeriojs/cheerio/blob/8ee743e89ee03f04bba971ade430490b079abdaf/.travis.yml#L2-L4). I confirmed locally, so it looks like this is specific to that version of the platform.

@fb55 I won't be able to look into this until tomorrow; in the mean time, do you have any idea what could be the problem?

blondyn:
Hi again. It seems that introduction of absolutize function in css-select is causing this. However I don't know the code well, might be something else. What is certain is that commit (fb55/css-select@0f0daad251094287a6223ba24467bf073dc2c0fe) is a cause for this fail.

Sorry if this is not appropriate place for reference issue from other library. I just though that cheerio might be feeding css-select api with some deprecated data.

jugglinmike:
Thanks! That looks to be a regression in CSSselect. I tried and failed to patch it, so I've settled for submitting a failing test case: https://github.com/fb55/css-select/pull/30 . Hopefully @fb55 will have better luck than I.

fb55:
Sorry for the late response. I still don't fully understand the error (haven't looked at the code in a while), for now just reverting 29d38e3 is probably the best choice.

Delgan:
Ok, so, I spent some hours trying to solve this issue, and while I almost reach the goal, something disturbs me.

I am not sure to fully understand the issue.

With reference to the unit test from  [PR #30](https://github.com/fb55/css-select/pull/30):
```javascript
"should query array elements directly when they have parents"

var dom = makeDom("<div id=foo><p>foo</p></div>")[0];
var ps = CSSselect("p", [dom]);
assert.deepEqual(CSSselect("p", ps), ps);
```

What happens here:
* Create the dom.
* Query the dom to get every `<p>` tag and store them into an array.
* Query this array to find `<p>` tag and then check it equals to the previous one found.

So, from an array `[ {p_tag} ]`, the query should return the same result `[ {p_tag} ]`... Why that?

[...]

PR #784, opened 2015-11-24
Update css-select to 1.2.0

fb55:
@Delgan fixed a long-standing bug in css-select (known from eg. #728).


jugglinmike:
Thanks;, @fb55!

Issue #780, opened 2015-11-14
Unexpected result while using .find(">") on a string containing <html> tag


Delgan:
Hello.

I had not had the opportunity to say it so far, but I wanted first of all to thank you for the wonderful tool that is Cheerio.js. You have done an amazing job, I can not express how much your work is valuable.
Really, thank you.

I recently encountered an unexpected behavior of Cheerio.js which I suspect to be a bug.

I opened [a question on StackOverflow](http://stackoverflow.com/questions/33713139/cheerio-and-find-return-too-many-elements-compared-to-jquery#) but I decided to come here and report this as an issue.

I try to parse HTML:
```html
<html>
    <head></head>
    <body>
        <div>
            <table>
                <tr>
                    <td>
                        <a href="/link_1.php">Link 1</a>
                    </td>
                    <td>
                        <a href="/link_2.php">Link 2</a>
                        <a href="/link_3.php">Link 3</a>
                    </td>
                    <td>
                        <a href="/link_4.php">Link 4</a>
                        <a href="/link_5.php">Link 5</a>
                    </td>
                </tr>
            </table>
        </div>
    </body>
</html>
```

Using this code:
```javascript
var cheerio = require("cheerio");
var $ = cheerio.load(html);
var page = $.root();

var tr = page.find("tr");

console.log(tr.find("> :nth-child(2) a").length);
```

What happens is that this code returns `5` whereas the second child of the `<tr>` tag contains only 2 links.

A test case can be found [here](https://jsfiddle.net/ub1wL25y/2/).
I also tried to [remove `$.root()`](https://jsfiddle.net/ub1wL25y/4/) and to [use `$.parseHTML()`](https://jsfiddle.net/ub1wL25y/5/) but the issue is still the same.

As you can see [here](https://jsfiddle.net/veajq33v/2/), jQuery works as expected.
Maybe is it related to a limitation of [selecting direct children using `querySelector()`](https://stackoverflow.com/questions/6481612/queryselector-search-immediate-children)?

What is strange is that removing `<html>` tag [seems to work](https://jsfiddle.net/ub1wL25y/6/), it returns `2`.
Moreover, `tr.find("> :nth-child(2)").find("a")` [works just fine too](https://jsfiddle.net/ub1wL25y/13/).

I really do not see what is wrong with my first code, so I opened this issue hoping you can help me and confirm whether or not it is a bug.

fb55:
[...]

Issue #742, opened 2015-08-24
.find('> bar') selects also nested descendants


algesten:
In this example I expect `> bar` to only select the *immediate descendant* from `foo`. But I get also the nested descendant. Cheerio 0.19.0

```xml
<foo>
    <bar>
        <bar></bar>
    </bar>
</foo>
```

```coffee
$ = require('cheerio').load('<foo><bar><bar></bar></bar></foo>')
$('foo').find('> bar').length
2   # I expect 1
```

fb55:
This is already fixed in the `css-select` repo, but triggered another bug that has to be fixed before the feature is available in cheerio.

clkao:
a release with this fix seems to depend on #728

algesten:
I wrote my own jquery-on-the-server partly because I got frustrated by this. https://github.com/algesten/zu

PR #796, opened 2016-01-02
Fix unexpected behavior while querying child

Delgan:
I followed [the solution of fb55](https://github.com/cheeriojs/cheerio/issues/780#issuecomment-166870067) and I added a unit test to ensure that #742 and #780 was correctly fixed.

### DIFF
--- a/lib/api/traversing.js
+++ b/lib/api/traversing.js
@@ -29,7 +29,9 @@ exports.find = function(selectorOrHaystack) {
     }, this));
   }
 
-  return this._make(select(selectorOrHaystack, elems, this.options));
+  var options = {__proto__: this.options, context: this.toArray()};
+
+  return this._make(select(selectorOrHaystack, elems, options));
 };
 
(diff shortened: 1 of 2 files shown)


jugglinmike:
This looks good to me. I'm not so hot on using `__proto__`, but there's precedence in the project for it, so I'll leave that discussion for another time.

@fb55 Did you author commit 4779762?

fb55:
@jugglinmike No, I did not. It follows a suggestion I made, @Delgan probably wanted to give credit (which I appreciate!). A mention in the commit message would be better IMHO, but I don't care enough about it to force a change.

LGTM otherwise!

jackwilsdon:
+1 on this, having the same issue.

adelgado:
This indeed fixes the issue, thanks :-)

I've seen elsewhere in the repo that releases are request-based, @fb55 is there anything missing for this PR to it be merged and a version containing the fix to be released? Thanks

PR #801, opened 2016-01-22
Add readme section in case for  using it with webpack

meandmax:
In the case you want to use it with the webpack output you need to configure webpack correctly in order to resolve the export in index.js right.


meandmax:
setup specific and not a general problem .... . :)

PR #808, opened 2016-01-31
Update lodash dependency to 4.1.0

leifhanack:
Lodash 4 has some braking changes. This PR reflects this and updates lodash dependency to 4.1.0.


fb55:
@leifhanack Thanks a lot for looking into this!

PR #809, opened 2016-01-31
ignore Intellij Idea project files

leifhanack:


jugglinmike:
It is not appropriate to ignore files generated Intellij within this project because it is not a dependency. I take it that you are using that editor, though, so you should look into using a global gitignore file--that will save you the hassle of ignoring the same files in every project to which you'd like contribute.

GitHub has a brief explanation here: https://help.github.com/articles/ignoring-files/

leifhanack:
Ah, ok!

PR #813, opened 2016-02-05
Few optimizations

marijaselakovic:
Hi,

I suggest the following two optimizations:

1) It is performance wise to use instanceof instead of toString class comparison, at least in the first place (it is implemented in almost all environments). I have made simple jsperf test to illustrate this: http://jsperf.com/instanceof-vs-tostring-2

2) String conversion with String constructor is more expensive than using the implicit string conversion. Since the input to String constructor in attributes.js, line 195 is always the number, it is safe to implicitly convert it to string. jsperf test: http://jsperf.com/implicitvsstringconstructor

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -192,7 +192,7 @@ var readData = function(el, name) {
 
       if (hasOwn.call(primitives, value)) {
         value = primitives[value];
-      } else if (value === String(Number(value))) {
+      } else if (value === Number(value)+'') {
         value = Number(value);
       } else if (rbrace.test(value)) {
         try {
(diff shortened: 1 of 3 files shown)


fb55:
Haha, this could have come from me 5 years ago.

The central problem with pull requests like this is that they slightly improve performance in places that don't really matter while worsening readability (which translates to maintainability). With a bit more exposition:

Switching from `Number(value)` to `value - 0` would probably yield another improvement inside V8, but the result, `(value - 0) + ""`, is much harder to read than `String(Number(value))`, especially for people not used to the JS type system. At the same time, Safari with JSCore doesn't care which variant is used and V8 might optimize for this as well in the next version.

For the second change, I'm *guessing* that `Object.prototype.toString` will be below 0.1% when profiling eg. the cheerio test suite with V8's tick processor, which is not worth the effort. As this is a wild guess, this PR should be accompanied by actual performance numbers.

Thank you for your effort, but I'm closing this. Don't feel discouraged from opening another PR when you spot notable performance improvements – or my intuition about the two handled cases is wrong!

PR #827, opened 2016-03-11
Added `.serialize()` support. Fixes #69

### LINKED ISSUE #69: .serialize()?

twolfson:
This was originally opened as #632 but @twolfson lost write access to the source repo and there was a merge conflict so we resolved it all in this PR

-------------------------

Now that #631 is implemented, we can add `.serialize()` support. Unfortunately, this comes with adding `$.param()`. `$.param` has a second half that `.serialize()` doesn't leverage (see its `else` statement). I am not sure whether we want to add this/test it, or drop it and make `$.param` internal only for now.

In this PR:

- Added `.serialize()` support http://api.jquery.com/serialize/
- Added tests for `.serialize()`

### DIFF
--- a/lib/api/forms.js
+++ b/lib/api/forms.js
@@ -2,7 +2,21 @@
 // https://github.com/jquery/jquery/blob/2.1.3/src/serialize.js
 var _ = require('lodash'),
     submittableSelector = 'input,select,textarea,keygen',
+    r20 = /%20/g,
     rCRLF = /\r?\n/g;
+
+exports.serialize = function() {
+  // Convert form elements into name/value objects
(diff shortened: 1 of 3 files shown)


twolfson:
This PR was initially opened in https://github.com/cheeriojs/cheerio/pull/632 over 1 year ago. Can we please land it?

fb55:
@twolfson Sorry for the delay – and thanks for implementing the feature :)

twolfson:
woot, thanks for landing this =)

PR #832, opened 2016-03-19
Use explicit extension for require package.json

tarjei:
I got an error Error: Cannot find module "./package" when using webpack and setting `resolve.extensions = ['', '.js', '.jsx']`. As package here is not going to have another extension than .json I see no downside in this change.


fb55:
Resolved in 7b59afbc7a6aa39376021593e011bf718350bcb4

PR #842, opened 2016-04-05
serializeArray should not ignore input elements without value attributes

rgladwell:
`serializeArray` ignores input elements without value attributes. The value attribute is optional for most types of input elements, however:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-value

In this case, input elements without a value attribute should be parsed by this function, and default to an empty string value.

### DIFF
--- a/lib/api/forms.js
+++ b/lib/api/forms.js
@@ -28,21 +28,21 @@ exports.serializeArray = function() {
       var name = $elem.attr('name');
       var val = $elem.val();
 
-      // If there is no value set (e.g. `undefined`, `null`), then return nothing
+      // If there is no value set (e.g. `undefined`, `null`), then default value to empty
       if (val == null) {
-        return null;
-      } else {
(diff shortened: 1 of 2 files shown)


rgladwell:
Any chance this could get merged?

jugglinmike:
Thanks for the patch! This change seems like an improvement to me, but I want to make sure we're tracking stable behavior in jQuery. I've opened a documentation pull request to lock down the functionality:

https://github.com/jquery/api.jquery.com/pull/1018

If that's accepted, I'll merge this right away.

jugglinmike:
That jQuery documentation patch was just accepted, so this is good to land. There were some conflicts, so I resolved those and merged to `master` at commit 692e301092a05241260cfb79f95bd635ca722a44. Thank you for your contribution, Ricardo (and sorry for the delay)!

PR #849, opened 2016-04-19
Add Chinese Translation

hongrunhui:
I wonder if it is possible to do so? ^.^
I hope that the Chinese people can access this document more easily.=.=


fb55:
I like this in general, but don't see how it can be maintained going forward: AFAIK none of the recurring contributors speak Chinese.

You could perhaps add a Wiki page and we link to it from the readme with an explicit note that it could be outdated.

fb55:
Closing in favor of #882

PR #853, opened 2016-04-27
remove jsdom dep & install jsdom in benchmark

sartrey:
I'm trying to remove useless dep:jsdom
Please review my changes ~

### DIFF
--- a/benchmark/install-jsdom.js
+++ b/benchmark/install-jsdom.js
@@ -1,0 +1,4 @@
+const exec = require('child_process').execSync
+
+console.log('try to install jsdom')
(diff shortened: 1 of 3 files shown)


fb55:
The reason for moving jsdom to `optionalDependencies` was its dependency `contextify`, which didn't compile on all node versions. Recent versions of jsdom don't require `contextify`, so it can be moved back to `devDependencies` (and the version can be increased to `^8.4.0`).

sartrey:
Fine! I'll close this pr and let's use devDependencies instead ~
Any plan to do this in a fix version, like 0.20.1?
I could hardly wait to use smaller cheerio ...

PR #855, opened 2016-04-27
Document, test, and extend static `$.text` method

jugglinmike:
As noted in the commit messages, this method was previously undocumented and untested. Add tests, and implement a change that, while technically backwards-incompatible, promotes parity between this method and the static methods `$.html` and `$.xml`.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -821,6 +821,13 @@ $.xml()
 //=>  <media:thumbnail url="http://www.foo.com/keyframe.jpg" width="75" height="50" time="12:05:01.123"/>
 ```
 
+You may also render the text content of a Cheerio object using the `text` static method:
+
+```js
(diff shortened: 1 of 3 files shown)


Readme.md
```diff
@@ -821,6 +821,21 @@ $.xml()
...
+$ = cheerio.load('This is <em>content</em>.')
+$.text()
+//=> This is content.
+```
+
+The method may be called on the Cheerio module itself--be sure to pass a collection of nodes!
+
+```js
+$ = cheerio.load('<div>This is <em>content</em>.</div>')
+cheerio.text($('div'))
```
fb55: There could perhaps be a test case for this as well. LGTM otherwise!

fb55:
Merged as 54359c92625c305f20fe765d54f7a4adde260892. Not sure why Github isn't showing it.

Issue #862, opened 2016-05-10
Update engine minimum to allow lodash removal


jsumners:
Currently, cheerio's package.json says it works with node.js 0.6 and later:

1. 0.6 is long unsupported
2. 0.10 and 0.12 will be unsupported in a few short months -- https://github.com/nodejs/LTS#lts_schedule

If the supported engines were updated to `>= 4.0.0` then there is nothing in the code base, that I have seen in a cursory read, that depends on lodash that is not achievable with native JavaScript. Fewer dependencies, and native code, is a good thing.

sotojuan:
Agreed—lodash makes up a big part of cheerio's filesize.

An alternative is to install the lodash functions from npm individually since cheerio only uses a few.

Would be more than happy to do a pull request for this.

fb55:
Regarding newer node versions: node 0.6 already has array extras that could be used for the functionality, but lodash is simply faster, as it ignores eg. sparse arrays.

@sotojuan A PR would be great :)

fb55:
Should be fixed with #864

PR #864, opened 2016-05-13
Use individual lodash functions

sotojuan:
May help with #862.

I replaced lodash with individual packages for each function. This brought down the size by ~4mb. Not too bad!

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -1,7 +1,4 @@
-var forEach = require('lodash.foreach'),
-    extend = require('lodash.assignin'),
-    some = require('lodash.some'),
-    $ = require('../static'),
+var $ = require('../static'),
     utils = require('../utils'),
     isTag = utils.isTag,
     domEach = utils.domEach,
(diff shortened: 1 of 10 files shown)


redwoodfavorite:
👍  Another consequence of this change is that lodash no longer pollutes the global namespace.  I am actually using the react testing library 'enzyme' and it's dependency on 'cheerio' is frustrating for that reason (currently overriding underscore)

Would like to see this change merged! :)

sotojuan:
Ping @fb55 @jugglinmike :-)

fb55:
IMHO keeping the `_` variable and creating an object inside each file would be a bit more readable.

sotojuan:
Hm, I can see that. I don't mind doing that—what does @jugglinmike think?

sotojuan:
Sorry for the late PR—been busy. I think I did what you meant, let me know.

fb55:
Awesome, thanks a lot :)

ianstormtaylor:
It seems like this change should be done using the `lodash/bind` syntax that `4.0` supports (which will cause an even greater reduction in size when using `cheerio` with other dependencies) instead of the `lodash.bind` syntax which requires pegging a module for each function.

sotojuan:
@ianstormtaylor if @fb55 agrees I don't mind submitting a PR for that. So it'd just be installing lodash fully again but using `var bind = require('lodash/bind')`? I'll have to test it later! Thanks.

ianstormtaylor:
@sotojuan yup, exactly like that! It should mean that other deps using `lodash` will also share the deps, even if they require the whole module.

sotojuan:
👍  We'll see what the repo owners think and I'll implement as needed.

billyjanitsch:
@sotojuan @ianstormtaylor I implemented the change you suggested in #913.

PR #878, opened 2016-06-17
Make text() non-recursive

sirwart:
I was having stack overflow issues so I made text() non-recursive

### DIFF
--- a/lib/static.js
+++ b/lib/static.js
@@ -114,14 +114,14 @@ exports.text = function(elems) {
   }
 
   var ret = '',
-      len = elems.length,
+      q = elems.slice(),
       elem;
 
-  for (var i = 0; i < len; i++) {
-    elem = elems[i];
+  for (var i = 0; i < q.length; i++) {
+    elem = q[i];
     if (elem.type === 'text') ret += elem.data;
     else if (elem.children && elem.type !== 'comment') {
-      ret += exports.text(elem.children);
+      Array.prototype.push.apply(q, elem.children);
     }
   }
 


lib/static.js
```diff
@@ -114,14 +114,14 @@ exports.text = function(elems) {
...
       elem;
 
-  for (var i = 0; i < len; i++) {
-    elem = elems[i];
+  for (var i = 0; i < q.length; i++) {
+    elem = q[i];
     if (elem.type === 'text') ret += elem.data;
     else if (elem.children && elem.type !== 'comment') {
-      ret += exports.text(elem.children);
+      Array.prototype.push.apply(q, elem.children);
```
leahciMic: This changes the behaviour and results in ordering errors. I'm not sure why it passed on Travis, and Travis won't allow me to view the build right now or sign in. I'll try checking again later.

It fails for me locally with:

```
  1) cheerio .text (cheerio object) : should return the text contents of the specified elements:
     Error: expected 'This is .content' to equal 'This is content.'
      at Assertion.assert (node_modules/expect.js/index.js:96:13)
      at Assertion.be.Assertion.equal (node_modules/expect.js/index.js:216:10)
      at Context.<anonymous> (test/api/utils.js:41:33)

  2) cheerio .text (cheerio object) : should include text contents of children recursively:
     Error: expected 'This is  text.a child with  followed by  and some finalanother child and  not a commentone last child' to equal 'This is a child with another child and  not a comment followed by one last child and some final text.'
      at Assertion.assert (node_modules/expect.js/index.js:96:13)
      at Assertion.be.Assertion.equal (node_modules/expect.js/index.js:216:10)
      at Context.<anonymous> (test/api/utils.js:51:33)

  3) cheerio .text () : should return the rendered text content of the root:
     Error: expected 'This is  text.a child with  followed by  and some finalanother child and  not a commentone last child' to equal 'This is a child with another child and  not a comment followed by one last child and some final text.'
      at Assertion.assert (node_modules/expect.js/index.js:96:13)
      at Assertion.be.Assertion.equal (node_modules/expect.js/index.js:216:10)
      at Context.<anonymous> (test/api/utils.js:56:27)
```

You would need to insert the children at the current index using `splice` or something.

leahciMic:
@sirwart how big is the node hierarchy you were calling `text()` on?

sirwart:
Hey Michael, I think the root cause of this issue is that the html parser is not handling closed <td> tags correctly. I was trying to parse a really large table without closed tags and it created a much more deeply nested structure than a web browser, which was then hitting the stack limit when trying to operate on it. Even though switching to a non-recursive solution would solve the stack limit problem, it wouldn't really solve the underlying problem (which isn't in cheerio) so I'm going to close this request)

leahciMic:
@sirwart no problems at all! if you end up creating an issue or pull-request on htmlparser2 please link to it here.

If you are able to share the HTML that caused this issue, I'd be more than happy to help take a look.

sirwart:
Here's the file if you're curious, it's from killedbypolice.net: https://www.dropbox.com/s/ad3tk8hjswqg7bu/KBP.html.

I ended up re-writing my scraper in Go though, so I'm not planning on spending more time on the node implementation.

Issue #762, opened 2015-10-06
ES6 arrow functions


gajus:
Might be worth mentioning somewhere in the documentation that cheerio will not work as expected with arrow-function.

Arrow function `this` value refers to the lexical `this` value (`this` in the scope of the declaration), i.e.

```js
$('div').each(() => {
    $(this).text(); // "this" does not refer to the element.
});
```

You can either continue to use the functional expression or use the parameter value, e.g.

```js
$('div').each((element) => {
    $(element).text(); // "this" does not refer to the element.
});
```

manjufy:
It did not work either of them.
I ended up doing

```
$('div').each(function(index, element) {
    $(element).text();
});
```

Rotimi-Best:
> that cheerio will not work as expected with arrow-function.
Really??? I never knew about that, it is actually in the docs that `this` or `element` could be used in the callback.

PR #879, opened 2016-06-17
make readme examples for ES2015 friendly

zeke:
`$(this)` will not work within "arrow" functions, where `this` refers to the lexical `this` value in the scope of the declaration.

This PR updates removes `$(this)` from the README in favor of `$(el)`, so the code examples can be rewritten in ES2015 and will still work.

cc @gajus, https://github.com/cheeriojs/cheerio/issues/762

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -415,8 +415,8 @@ Iterates over a cheerio object, executing a function for each matched element. W
 ```js
 var fruits = [];
 
-$('li').each(function(i, elem) {
-  fruits[i] = $(this).text();
+$('li').each(function(i, el) {
+  fruits[i] = $(el).text();
 });
 
 fruits.join(', ');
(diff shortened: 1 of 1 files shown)


zeke:
whoa, @coveralls 🏇

tomByrer:
Plus `el` reads much better than `this`.

gajus:
> This PR updates removes $(this) from the README in favor of $(el), so the code examples can be rewritten in ES2015 and will still work.

While you are at it, I'd change use of regular functions to arrow expressions.

zeke:
Bump.

I still think this PR is worth landing, for the sake of making the readme friendly for users of arrow functions.

jugglinmike:
This change makes the documentation less expressive overall because the `this`
value of those functions is no longer demonstrated. It's beyond the scope of
this project to help folks learn to use new language features; if consumers use
arrow functions, we have to assume they understand the implications that has on
their code and any examples they find elsewhere. The second parameter is already
listed here (and given the conventional name `el`), so I think that should be
sufficient.

Issue #875, opened 2016-06-14
$(undefined).prop('href') throws error


rla:
`$(undefined).prop('href')` throws:

```
TypeError: Cannot read property 'hasOwnProperty' of undefined.
```

With jQuery the same expression returns `undefined`.

Tested with Cheerio 0.20.0

sotojuan:
Seems like the problem is [here](https://github.com/cheeriojs/cheerio/blob/master/lib/api/attributes.js#L94)? Seems an easy fix.

rla:
@sotojuan, if you are interested in taking over then go ahead.

elias-winberg:
It should be relatively easy to fix, no? If `el` is `undefined` immediately return `undefined`. I'll try to create a pull request.

Delgan:
Is it related to https://github.com/cheeriojs/cheerio/issues/554 ?

elias-winberg:
Sort of, jQuery creates the element even if the attribute value is undefined, and cheerio doesn't.

rla:
I add some background info.

The bug might also be triggered when the selector does not match, like in the code:

```js
$('.not-exists').attr('href')
```

It was not enough to check that `$('.not-exists')` is truthy before calling `.attr()` method. $ is Cheerio here. I ported the code from code using querySelector that behaves way more logically than jQuery.

Looking through the current Cheerio tests, it does not seem to consider negative cases when some input is undefiend or null.

elias-winberg:
`.attr` seems to be working as expected though (only `.prop` throws an error if the selector is invalid). I think this is due to the fact that the [`.getProp`](https://github.com/cheeriojs/cheerio/blob/e7d18af05ff1d23f17d858590e2e1d299f7bdf75/lib/api/attributes.js#L93) function lacks this [line](https://github.com/cheeriojs/cheerio/blob/e7d18af05ff1d23f17d858590e2e1d299f7bdf75/lib/api/attributes.js#L31).

fb55:
Fixed in https://github.com/cheeriojs/cheerio/pull/880

PR #880, opened 2016-06-17
.prop now returns undefined if given an invalid element or tag (#875)

elias-winberg:
Fix for issue https://github.com/cheeriojs/cheerio/issues/875.


leahciMic:
👍 for this, I have the same issue.

fb55:
Thanks!

PR #887, opened 2016-07-04
Add baseHref option to cheerio.load, and have prop() resolve urls

leahciMic:
This pull request adds a baseHref option to cheerio.load. When this option is set, .`prop(attrName)` will resolve `src` and `href` attributes as jQuery would relative to the baseHref.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -160,7 +160,8 @@ of the default parsing options:
 ```js
 $ = cheerio.load('<ul id="fruits">...</ul>', {
     normalizeWhitespace: true,
-    xmlMode: true
+    xmlMode: true,
+    baseHref: 'http://example.com/'
 });
 ```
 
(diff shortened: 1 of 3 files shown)


leahciMic:
@matthewmueller I wasn't quite sure how to add this functionality. Basically in the browser, and in jQuery when accessing a property such as `src` on an image or `href` on an anchor, you'll receive an absolute url. This pull-request intends to give cheerio this same behaviour.

Also, `cheerio.load` takes options as a second argument to pass through to `htmlparser2`. I'm not sure if it's a good idea to tack on the `baseHref` option to that object. Your thoughts would be appreciated.

fb55:
This is IMHO the right approach for this and I'd be fine with adding it.

sjungwirth:
Any update on this? I would like this functionality as well.

PR #898, opened 2016-07-28
If null is passed to html, empty the element rather than throwing. 

### LINKED ISSUE #897: passing null to .html() should not throw
Currently if you pass null to html, it throws:

```
     TypeError: Cannot read property 'cheerio' of null
      at null.<anonymous> (lib/api/manipulation.js:378:22)
      at exports.domEach (lib/utils.js:52:24)
      at exports.html (lib/api/manipulation.js:373:3)
      at Context.<anonymous> (test/api/manipulation.js:1350:15)
```

I believe it should empty the element, as jQuery does. PR to follow.

dhendo:
Fixes #897

### DIFF
--- a/lib/api/manipulation.js
+++ b/lib/api/manipulation.js
@@ -367,6 +367,10 @@ exports.html = function(str) {
     if (!this[0] || !this[0].children) return null;
     return $.html(this[0].children, this.options);
   }
+  
+  if(str === null){
+    str = '';
+  }
 
   var opts = this.options;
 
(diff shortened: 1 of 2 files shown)


jugglinmike:
Thanks for the patch! Unfortunately, the reported behavior in jQuery is neither [documented](https://api.jquery.com/html/) nor [explicitly tested](https://github.com/jquery/jquery/blob/3e3b09d68d317a3fe567e377e9032319ac5e0829/test/unit/manipulation.js) by that project. That means this isn't the kind of incompatability we're generally concerned with. If anyone would like to work with the jQuery project to make this behavior part of their formal API, then please report back here once you're done.

For now, though, the best practice (in Cheerio or jQuery) would be to use [the `empty` method](https://api.jquery.com/empty/) for this case.

PR #913, opened 2016-08-25
Use modular lodash package

billyjanitsch:
As of lodash 4.14.0, individual packages (e.g. `lodash.foreach`) are zero-dependency modules (see the [changelog](https://github.com/lodash/lodash/wiki/Changelog#v4140)), meaning that if you install several of them, they do not share any code.

This completely negates (and actually worsens) the bundle size savings introduced in #864, since cheerio now depends on so many individual method packages. The solution is to use the modular version of lodash (installing lodash and importing from `lodash/method` rather than from the `lodash` monolith) as mentioned [here](https://github.com/cheeriojs/cheerio/pull/864#issuecomment-234014347).

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -8,9 +8,9 @@ var $ = require('../static'),
     rspace = /\s+/,
     dataAttrPrefix = 'data-',
     _ = {
-      forEach: require('lodash.foreach'),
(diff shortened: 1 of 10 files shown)


ianstormtaylor:
This is awesome!

fb55:
Thanks a lot!

PR #926, opened 2016-10-03
Update hasClass method

sufisaid:
If function is called with empty string it cause infinite loop.


fb55:
Could you provide a test case?

jugglinmike:
Squashed and merged to `master` at commit 9040598aa16b95cd5d6c9c454f28d0dbc0a6529b. Thank you!

Issue #936, opened 2016-11-01
last() should take an optional argument


rightaway:
`last()` gets the last element, but it would be good if it could take an optional argument, where something like `last(2)` would get the second-last item.

fb55:
As noted in #939, this is not part of the jQuery API and will therefore not be supported.

PR #939, opened 2016-11-14
Added an optional argument for last()

Humad:
Addressed https://github.com/cheeriojs/cheerio/issues/936

Last takes an optional argument so that something like last(2) will return the second-to-last item


fb55:
This isn't part of the jQuery API, so we won't support it (but we would support any equivalent method that is part of the jQuery API). Also, a test case and an entry in the readme would be helpful for future pull requests :)

PR #945, opened 2016-11-23
#944 fixing missing prop(‘outerHTML’) implementation.

### LINKED ISSUE #944: .prop('outerHTML') returns undefined
cheerio:
`require('cheerio').load('<div></div>')('div').prop('outerHTML') // undefined`

jQuery:
`$('<div></div>').prop('outerHTML') // '<div></div>'`

i have implemented a fix which adds a switch case for 'outerHTML' in the exports.prop function within lib/attributes.js which uses the following implementation: `this.clone().wrap('<container />').parent().html()`

I will make a pull request, please let me know your thoughts.

bill-bishop:
fixes #944
Added an ‘outerHTML’ case to the switch in the prop function, which wraps a clone of `this` in a container element, and sets `property` to that container's HTML

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -125,6 +125,9 @@ exports.prop = function (name, value) {
       case 'nodeName':
         property = this[0].name.toUpperCase();
         break;
+      case 'outerHTML':
+        property = this.clone().wrap('<container />').parent().html();
+        break;
       default:
         property = getProp(this[0], name);
     }
--- a/test/cheerio.js
+++ b/test/cheerio.js
@@ -350,5 +350,14 @@ describe('cheerio', function() {
         expect($b('div').foo).to.be(undefined);
       });
     });
+
+    describe('.prop', function () {
+      describe('("outerHTML")', function () {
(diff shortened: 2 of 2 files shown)


gajus:
Same for `innerHTML`, `className`, etc. :-(

gajus:
@fb55 has this been done on purpose?

fb55:
I would think that nobody needed it, so it was never implemented. Patches always welcome (even though it might take some time for them to be reviewed).

balint42:
tested: works: 👍

william-mcmillian:
@gajus `prop('innerHTML')` and `prop('className')` can be replaced with `.html()` and `.attr('class')` but if there's other properties not supported i'd be glad to fix those as well

fb55:
Thanks a lot @bill-bishop, merged this into the v1.0.0 branch!

jcrben:
@fb55 @bill-bishop someone should close the associated issue https://github.com/cheeriojs/cheerio/issues/944

fb55:
Thanks for the notice @jcrben!

PR #947, opened 2016-11-25
let apple = $('#apple');$('#fruits').add(apple);

1003657663:
the right result is #fruits has apple
but I found apple has fruits

I don't need $('#fruits') = $('#fruits').add(apple);//and apple is wrong
I need $('#fruits).add(apple); then that fruits is changed;
I changed this bug

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ lib-cov
 cover_html
 .c9revisions
 coverage
+.idea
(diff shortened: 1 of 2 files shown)


fb55:
This method is [supposed to](https://api.jquery.com/add/):

> Create a new jQuery object with elements added to the set of matched elements

I'm pretty sure that you try to use it for something else. Otherwise, could you add a test case?

PR #949, opened 2016-11-28
a.add(b) is wrong

1003657663:
a.add(b)
result should be b is added into a,I fixed this wrong


fb55:
Whitespace is off, and needs a test case.

Issue #866, opened 2016-05-17
Non-latin characters get HTML-encoded with decodeEntities=true


dsavenko:
Hi,

Consider this code:

``` javascript
var cheerio = require('cheerio')
var ch = cheerio.load('<div>абв</div>', { decodeEntities: true })
console.log(ch('div').html())
```

It prints `&#x430;&#x431;&#x432;`. If I set `decodeEntities` to `false`, the output will be the expected `абв`. 

Two issues here:
1. This is not how `decodeEntities` is supposed to work. I tested with `htmlparser2` directly, and it works as expected both ways (the code is below).
2. Htmlparser recommends always set `decodeEntities` to `true` for security reasons. 

Test code for htmlparser:

``` javascript

var htmlparser = require("htmlparser2");
var parser = new htmlparser.Parser({
    ontext: function(text){
        console.log("-->", text);
    }
}, {decodeEntities: true});
parser.write("<div>абв</div>");
parser.end();
```

Output: `--> абв` (as expected)

P.S. Cheerio version: 0.20.0

shelldweller:
This was reported in #565 and for some reason closed. I agree, `.html()` should not escape non-ASCII characters to entities, but should return Unicode string, like `.text()` does.

davidbayo10:
This code is working for me. The problem was html response which was not encoding correct

``` javascript
const request = require('request');
const cheerio = require('cheerio');
const iconv = require('iconv-lite');

function getDOM(url) {
  return new Promise(function (resolve, reject) {
    request({ uri: url, encoding: null }, function (err, res, html) {
      if (err) {
        reject(err);
      } else {
        html = iconv.decode(html, 'ISO-8859-1');
        resolve(cheerio.load(html));
      }
    });
  });
}
```

Cactusbone:
looks like it's manually done that way.

```
 if (opts.decodeEntities && !(elem.parent && elem.parent.name in unencodedElements)) {
    data = entities.encodeXML(data);
  }
```

I suggest adding a parameter `encodeEntities`, defaulting to `decodeEntities` value, which can be overriden via options.

Cactusbone:
found a way around the problem : 

```js
var cheerio = require('cheerio')
var ch = cheerio.load('<div>абв</div>', { decodeEntities: true })
console.log(ch('div').html({ decodeEntities: false }))
```

rlidwka:
@Cactusbone :

I stumbled across this issue, so I took my time to test this workaround:

[...]

PR #953, opened 2016-11-30
Decode entities workaround test

Cactusbone:
workaround for bug #866 
test to ensure workaround will work :)


fb55:
This option won't work in v1.0.0

PR #976, opened 2017-02-12
should not alter quotes in attribute values

koenpunt:


fb55:
We could be smarter about this, but it's a separate issue.

koenpunt:
Separate of what?

fb55:
Sorry, poor phrasing: This would be an issue for dom-serializer, or we can switch to parse5's serializer.

Issue #977, opened 2017-02-12
Builds succeed even if test don't pass


koenpunt:
As can be seen at #976, where the commit is marked with a check. But if you follow the link to Travis, you'll see that the tests did not pass: https://travis-ci.org/cheeriojs/cheerio/jobs/200843828

jugglinmike:
Nice catch, and thanks for the report! I've opened gh-978 to address the issue.

fb55:
Fixed in #978

PR #978, opened 2017-02-12
Report test failures in CI

jugglinmike:
Due to occasional unavailability in the code coverage reporting service,
the continuous integration environment does not reject builds that
experience issues with this task.

This behavior was previously implemented in such a way to allow errors
of *any* kind--including test failures. Re-implement to limit the scope
of "allowed errors," to the reporting command only.


jugglinmike:
This should resolve gh-977.

fb55:
Nice!

Issue #700, opened 2015-04-13
Is cheerio still 8x faster than jsdom?


DavidBruant:
This part of the readme has been written 3,5 years ago.

https://github.com/cheeriojs/cheerio/commit/ba80a89428a44f31759e5fe009be731556c4a3b9

Is it still the case (especially regarding the 4.x serie of jsdom)?

fb55:
jsdom 4.x switched the HTML parser to a slower & more accurate one and now uses io.js' `vm` module instead of contextify, which isn't a speed improvement over cheerio (since cheerio doesn't evaluate JS at all). The actual comparison depends on the workload, cheerio should definitely be faster in pretty much all cases, though.

domenic:
I'd like this to be reopened and reinvestigated. jsdom doesn't use contextify or vm anymore for the simple types of workloads that cheerio supports. It is also notably faster on selector-based traversal. We'd like to get the benchmark from which the "8x faster" number comes from open-sourced so that jsdom can work on improving itself. If that's not possible, toning down the language to "sometimes faster than jsdom (no benchmarks though)" or similar would be more sensible ;).

fb55:
The claim should definitely be substantialized. I assume the number originates from @matthewmueller's [screencast](https://vimeo.com/31950192), but he might have more details. In general, cheerio's [benchmarks](https://github.com/cheeriojs/cheerio/tree/8ee743e89ee03f04bba971ade430490b079abdaf/benchmark) also run on jsdom, with varying differences (I haven't run them recently).

On the other hand, we might want to highlight that cheerio runs on pretty much every JS platform, not only on io.js and node, but also on Rhino/Nashorn, inside browsers, you name it.

tconroy:
Hey folks, 
curious if anyone had taken recent benchmarks of cheerio vs jsdom performance? Trying to decide which to use for a project that utilizes lots of server-side node parsing.

matthewmueller:
yah, the "8x faster" claim was correct about 4 years ago. I have no idea about now and would happily accept a PR that these metrics over time.

cmawhorter:
Been looking for some numbers myself.  FWIW -- Came across [this page](https://habrahabr.ru/post/163979/) which seems to substantiate the claim (and then some), but I didn't dig all that deeply.  There is a repo: https://github.com/seriyps/html-parsers-benchmark

domenic:
I don't read Cyrilic that well but that post looks to be from 2012 to me.

[...]

PR #985, opened 2017-02-21
Use parse5 as a default parser (closes #863)

### LINKED ISSUE #126: Parse failure
Cheerio seems to fail to parse this HTML: http://pastebin.com/iEurSsVW obtained from http://www.politico.com/

I believe the problem lies with the javascript in the script tag starting at line 1575. The JS  contains an embedded script tag itself, which may be badly escaped. I'm not very experienced with this style of HTML, perhaps this is a WontFix for cheerio?

That parsing has failed is shown by the fact that queries like `$('#secondaryContent')` are successful when referring to elements prior to line 1575 but not afterwards (e.g. `$('#popularMostRead')`).

Thanks very much for cheerio, I switched to it yesterday when I couldn't get jsdom to run on heroku.

### LINKED ISSUE #240: Recusion in parse.js: Maximum call stack size exceeded
On certain HTMLs I receive the following error.  Maybe connect() could be implemented without a recursion?

```
(...)/node_modules/cheerio/lib/parse.js:47
  _.each(dom, function(elem) {
                      ^
RangeError: Maximum call stack size exceeded
```

### LINKED ISSUE #405: Loading https://support.microsoft.com/ using Cheerio dom manipulation not working properly.
Load the website https://support.microsoft.com/ html using cheerio. 

``` js
 var $ = cheerio.load(body); // body is html of this above url view-source:https://support.microsoft.com/
```

Adding a base tag and removing the same does not works. 

``` js
var baseTag =`<base id="tag-1" href="' + someUrl + '">`
Adding :  $('head').append(baseTag);
Removing :  $('#tag-1').remove();
```

Above codes works every time but not for this html.

### LINKED ISSUE #422: <td> tags with no <tr> are not properly handled
I am being passed some broke-ass-html where there is a missing `<tr>` tag before a `<td>`. This causes my queryies for table rows to fail as cheerio (or maybe the htmlparser) is not scoping the `<td>` inside of a logical `<tr>` as the browser does.

``` js
var cheerio = require('cheerio');
var str = '<table><td>bar</td></tr></table>';
var $ = cheerio.load(str);
console.log($.html());
```

### LINKED ISSUE #522: Less-than character in text node breaks parser
Consider this HTML:

``` html
<!DOCTYPE HTML PUBLIC "-//W3C//DTD HTML 4.01 Transitional//EN" "http://www.w3.org/TR/html4/loose.dtd">
<html>
<body class="flat_page">

<div class="pagecontent ">

    <div class="maincontent">


        <div class="badge_row depressed badge_gamecard_page">
            <div class="badge_row_inner">
                <div class="badge_detail_tasks">
                                                                                            <div class="badge_card_set_card unowned">
                            <div class="badge_card_set_text">
                                                                <3LuV<3                             <div style="clear: right"></div>
                            </div>
                        </div>
                                                                                            <div class="badge_card_set_card unowned">
                            <div class="badge_card_set_text">
                                                                Old Dude > Sen'SeY                              <div style="clear: right"></div>
                            </div>
                        </div>
                </div>
                            </div>
        </div>
    </div>
</div>

</body>
</html>
```

The first text node doesn't get parsed correctly. For example, the following code:

``` js
var cheerio = require('cheerio')

var body = require('fs').readFileSync('shit.html')
var $ = cheerio.load(body);
console.log($('.badge_card_set_text').eq(0).text().trim()); 
console.log($('.badge_card_set_text').eq(1).text().trim()); 
```

Outputs this:

```

Old Dude > Sen'SeY
```

The provided HTML is a simplified version of [this](http://steamcommunity.com/id/zombine/gamecards/259830/).

### LINKED ISSUE #694: incorrect html4 parsing
``` javascript
console.log(cheerio.load('<html><body><li><ul>item1<ul>item2<ul>item3').html());
```

shows

``` html
<html><body><li><ul>item1<ul>item2<ul>item3</ul></ul></ul></li></body></html>
```

should be

``` html
<html><body><li><ul>item1</ul><ul>item2</ul><ul>item3</ul></li></body></html>
```

### LINKED ISSUE #746: Unexpected behaviour with custom tag "line"
``` javascript
cheerio.load("<line>text</line>").html()
```

returns `"<line></line>text"`, expected `"<line>text</line>"`. The parsed tree is wrong, it's not just the html output bug.

### LINKED ISSUE #826: Cheerio failing with lax HTML structure?
I have a page which I am trying to scape of the form:

```
<tr><td><a href="...">mylink</a></td></tr>
<tr><td><a href="...">mylink</a></td></td></tr>
<tr><td><a href="...">mylink</a></td></td></tr>
```

Which seems to cause issues with the following logic:

```
var nodes = $('div table[cellpadding=3] tr td a')
```

The result here is one match, as opposed to three. I didn't see anything in the cheerio docs on how to normalise the HTML, so can anyone indicate whether this is something where depend on another library to normalise or whether cheerio should deal with it. Note, if I take out the extra </td> then it returns the right number of anchor tags.

Note, that JQuery handles this in Google Chrome, but it would appear this is probably because browsers normalise the HTML prior to rendering?

### LINKED ISSUE #860: Greater than followed by equal sign is lost
$.load('<=').html()
returns empty, instead of '<='

Now, this is important to me because I'm parsing html which can express "_A_ <= _B_" (A less or equal than B) as:
`<i>A</i><=<i>B</i>`

`$.load('<i>A</i><=<i>B</i>') == '<i>A</i><=<i>B</=<i>'`

This issue [#103](https://github.com/cheeriojs/cheerio/issues/103) had been issued for only greater sign. Can this be fixed too?

### LINKED ISSUE #863: Make it possible to use parse5 as optional(?) parser
Today I'm deprecating [whacko](https://github.com/inikulin/whacko), so, if cheerio maintainers don't mind, I'll send those who is interested in this feature here.

### LINKED ISSUE #907: Fails to parse datauris properly
If you have an element like this:

``` html
<div style='font-family:"butcherman-caps"; src:url(data:font/opentype;base64,AAEA...);'></div>
```

Parsing this element with cheerio will not return a correct value for the css property `src`:

```
$('div').css('src') -> url(data:font/opentype
```

I assume the css parser believes that the ; after opentype signals the end of the style property. The parser will need a special parser for handling datauris.

### LINKED ISSUE #915: Incorrect innerHTML of <xmp> tag
It seems that content of `<xmp>` tag should not be parsed at all. 
## Code

``` js
> var $ = require('cheerio')
undefined
> $('<xmp><h2></xmp>').html()
'<h2></h2>'
```
## Current output

``` js
'<h2></h2>'
```
## Expected output

``` js
'<h2>'
```

In browser the output is `'<h2>'` (http://codepen.io/anon/pen/VjyNQN inside console)
Jsdom seems to also return this output (https://tonicdev.com/57c2990a1ac3301400487da5/57c2990a1ac3301400487da6)

### LINKED ISSUE #937: text function duplicates characters when it ends in &# ampersand hash
```javascript
cheerio("<p>z&#</p>").text()
// => "zz&#
```

Notice that two z's end up in the return value of text.

This probably seems like a weird use case but this came about because we have a bunch of tests that generate random strings using http://chancejs.com/#string and it seems to create a string like this once per week during a test run which causes an assertion using .text() to fail.

I ran the same code with real jQuery and it behaves as expected:

```javascript
$j("<p>z&#</p>").text()
// => "z&#"
```

### LINKED ISSUE #971: childCombinator does breaks when there is an unclosed tag
Example:

```js
const input = `<!DOCTYPE html>
<head>
</head>
<body>
  <div id="headerInner">
    <h3>Welcome to WTW Cinemas</h2>

    <ul>
      <li>a</li>
      <li>b</li>
      <li>c</li>
    </ul>
  </div>
</body>
</html>`;

console.log( cheerio.load(input)('#headerInner > ul > li').length );

```

This produces output `0`. Expected output is `3`.

Example:

```js
const input = `<!DOCTYPE html>
<head>
</head>
<body>
  <div id="headerInner">
    <ul>
      <li>a</li>
      <li>b</li>
      <li>c</li>
    </ul>
  </div>
</body>
</html>`;

console.log( cheerio.load(input)('#headerInner > ul > li').length );

```

This produces output `3`. Expected output is `3`.

### LINKED ISSUE #997: pre tag parsing bugs
Raw html like so
```html
<pre>
A <- factor(A, levels = c("c","a","b"))
</pre>
```

will be loaded as:

```html
<pre>
A <- factor(a,="" levels="c("c","a","b"))" <="" pre="">
```

inikulin:
This PR  makes `parse5` a default parser for cheerio. If any parsing options are set to non-default values cheerio fallbacks to `htmlparser2`. Additionally, `useHtmlParser2` option was introduced to force usage of `htmlparser2`.  

- closes #863 
- closes #860 
- closes #126 (seems like was already fixed in htmlparser2, btw)
- closes #240 (not related to this issue, but was fixed by removing recursive node decoration)
- closes #405 
- closes #422 
- closes #522 
- closes #694 (wasn't a bug, htmlparser2 handled it correctly)
- closes #746 
- closes #826 
- closes #907 
- closes #915 
- closes #937 
- closes #971 
- closes #997


inikulin:
Please, do not merge yet. I'll go through open issues and reference the ones that will be closed by this PR along with regression tests.

jugglinmike:
This is awesome, thank you! I will give it a look as soon as I am able.

tommaton:
Hey guys, is there any timescale on when this will be merged in as would really like to start using parse5 with cheerio in a new project.

jugglinmike:
Not currently, no. Please remember that this is a non-trivial change and calls for careful review.

Readme.md
```diff
@@ -164,14 +164,15 @@ $ = cheerio.load('<ul id="fruits">...</ul>', {
 });
 ```
 
-These parsing options are taken directly from [htmlparser2](https://github.com/fb55/htmlparser2/wiki/Parser-options), therefore any options that can be used in `htmlparser2` are valid in cheerio as well. The default options are:
+These parsing options are taken directly from [htmlparser2](https://github.com/fb55/htmlparser2/wiki/Parser-options), therefore any options that can be used in `htmlparser2` are valid in cheerio as well. If any of these options is set to non-default value cheerio will implicitly use `htmlparser2` as an underlying parser. In addition, you can use `useHtmlParser2` option to force cheerio use `htmlparser2` instead of `parse5`. The default options are:
 
 ```js
 {
     withDomLvl1: true,
```
fb55: Is `withDomLvl1` implemented in the parse5 tree adapter?

lib/parse.js
```diff
@@ -12,20 +15,33 @@ exports = module.exports = function(content, options) {
...
+  root.parent = null;
 
   // Update the dom using the root
   exports.update(dom, root);
 
   return root;
 };
 
+function parseWithParse5 (content) {
+  var parseAsDocument = /^(\s|<!--.*?-->)*?<(!doctype|html|head|body)(.*?)>/i.test(content),
```
fb55: This is a smart hack, I'm not sure if it's expected though. IMHO we should only parse fragments right now and switch to full parsing in cheerio@1.0

inikulin: If we'll always use fragment parsing parser will always omit `<html>`, `<body>` and `<head>` tags and doctypes, so it will basically will return empty AST for full documents. On the other hand, enabling full parsing will always add these tags. Even in 1.0 you will still need this hack.

inikulin: Speaking clearly it just enabled the behavior that we currently have in cheerio. Unless we'll not add explicit API for document and fragment parsing, but as far as I can tell it will lead to breaking change.

fb55:
Okay, I didn't realize that's the case. My preferred final behavior would
be to have .load load documents and $() parse fragments. That suggestion
was only to keep backwards  compatibility for now.

Given that that won't be possible, we might consider a semver major bump
for this. At the same time, we could also adopt the parse5 DOM structure
and add hooks for the old DOM structure (getters for type, name etc.) for
backwards compatibility, with eventually an option to turn it off for
better performance.
-- 
– Felix

inikulin:
@fb OK, let's stick with this plan:
- [ ] use full-parsing for `.load()`
- [ ] use fragment parsing for `$`
- [ ] switch to parse5 AST
- [ ] add optional node decorators for backward compatibility

inikulin:
I just realized that it will not be possible to switch to parse5 AST, since we use htmlparser2 for XML.

jugglinmike:
Let me start be re-iterating: this is really great work! You've clearly done your homework when it comes to Cheerio's known bugs and backwards compatibility.

This is a significant change to the core behavior of the library. While the value is clear, it's also a risk. Our test suite is strong, but I'm not convinced it is thorough enough to identify all possible regressions.

It's also kind of surprising that any change to the parsing options triggers the use of a completely different parser, especially when one of the options is
`useHtmlParser2`. For instance:

```js
$ = cheerio.load('<ul id="fruits">...</ul>', {
    xmlMode: true,
    useHtmlParser2: false
});
```

It's not at all clear that example would actually use `htmlparser2`. It would be nice to use `parse5` for XML also, but [I can see why that's not possible](https://github.com/inikulin/parse5/issues/77).

I agree with @fb55's thoughts about releasing this as part of a new major version for Cheerio. That would:

- shield most consumers from regressions, unlikely as they seem (woe to those that use the `*` version specifier)
- give us the opportunity to make a "clean break" from the unfortunate "implicit document fragment" strategy we currently follow
- allow us to introduce a more intuitive parser configuration object. I'm open to suggestion about how this should look, but maybe something like this:

  ```json
    {
      "useHtmlParser2": {
        "withDomLvl1": true,
        "normalizeWhitespace": false,
        "xmlMode": false,
        "decodeEntities": true
      }
    }
  ```

Does that sound good to you two?

Readme.md
jugglinmike: I had the same question, so [I experimented a bit](https://gist.github.com/jugglinmike/3661b713ffb0393c5dba16375255389d) and found that Dom Level 1 support is analogous. [Here's the relevant source code](https://github.com/inikulin/parse5/blob/969ed22f53db38c101449e05a1b804ea456d4e04/lib/tree_adapters/htmlparser2.js#L7-L62). Great!

inikulin:
OK, I've just implemented changes we've discussed with @fb55 before: `$.load` always parser in document mode, while `$` is used for fragment parsing. 

Regarding @jugglinmike's suggestion about parser options: I agree that implicit switching of options doesn't look good. I like the idea of `useHtmlParser2` options object. But I'm not sure about moving `xmlMode` option in it, though. It's commonly used it it not quite intuitive to search it in `useHtmlParser2` object. Also, setting `xmlMode` to `true` implies that parsing behavior will be changed. I believe it would be better to not care about underlying parser when they use this option.

jugglinmike:
Good point. I don't want to let difficulty choosing names determine what
features we support, but I'm starting to think that this challenge is actually
a result of feature creep.

Do we really want to commit to simultaneously supporting both parsers for HTML
documents? Instead of allowing the user to control `xmlMode` *and* switch
between `parse5` and `htmlparser2`, maybe it makes more sense to just say,
"`parse5` is always used for HTML documents, and `htmlparser2` is always used
for XML documents."

This would certainly simplify configuration (basically, `xmlMode` could be
`true` or an object of options for `htmlparser2`). But while that resolves the
issue we're currently discussing, I think it might be a wiser choice for the
long-term maintenance of the project. That's because a hard delineation between
document parsers would tend to limit the variation in future bug reports. We
won't have to ask, "which HTML parser are you using?" and users won't have to
wonder the same thing when debugging their own code.

@fb55 I'd love your feedback here, too. Does it seem wise to support a single
parser for each document type? Or am I getting carried away by a preoccupation
with a minor detail about the options object?

inikulin:
Any feedback on this?

fb55:
Sorry for the delay :/

As far as I remember the main reason to keep htmlparser2 was to have a parser that does not modify the DOM structure, or supports weird DOM structures (self-closing tags!). I would agree that that's in the realm of feature-creep. For XML, [isaacs/sax-js](https://github.com/isaacs/sax-js) is the better option, although it is stricter with what it considers valid XML. (It's also what jsdom is using for xml.)

I'm not sure how we handle the DOM transition best. One option would be to continue to support the old names (ideally with the option to disable it), another to provide a [codemod](https://github.com/facebook/jscodeshift) that tries to rename all property accesses. For the latter option we could just stick with parse5's DOM tree and continue to use `withDomLvl1` in htmlparser2.

inikulin:
To conclude, let's figure out features which should get into this PR and thus in `1.0`. I believe we should use `options` structure proposed by @jugglinmike in https://github.com/cheeriojs/cheerio/pull/985#issuecomment-284250164, but make `xmlMode` top-level option. Regarding tree format and switch to sax parser I'm not quite sure if it should get into release. Maybe it makes sense to do changes iteratively? Or we'll need to make major version bump once again after tree format switch?

jugglinmike:
> (@inikulin) I believe we should use `options` structure proposed by @jugglinmike in https://github.com/cheeriojs/cheerio/pull/985#issuecomment-284250164, but make `xmlMode` top-level option.

Actually, I'm not in favor of this modification to my recommendation. This is
what I was describing above when I wrote, "It's also kind of surprising that
any change to the parsing options triggers the use of a completely different
parser [...]" and it also puts us "on the hook" for supporting two different
HTML parsers at the same time (something else I argued against earlier in this
conversation). What I'd like to see is simply:

    { xml: true }

or 

    { xml: { /* htmlparser2 options */ } }

Although we may end up removing the second form prior to releasing this.

> (@inikulin) Regarding tree format and switch to sax parser I'm not quite sure
> if it should get into release. Maybe it makes sense to do changes
> iteratively? Or we'll need to make major version bump once again after tree
> format switch?

This sounds good to me. You've done a great job here, but I don't want to keep
you "on the hook" for even more. Let's plan to land this in a `v1.0` feature
branch. Then @fb55 and I (and anyone else, including yourself) can iterate on
top of it.

...and on that note:

> (@fb55) I'm not sure how we handle the DOM transition best. One option would
> be to continue to support the old names (ideally with the option to disable
> it), another to provide a codemod that tries to rename all property accesses.
> For the latter option we could just stick with parse5's DOM tree and continue
> to use withDomLvl1 in htmlparser2.

This may be callous of me, but I'm not particularly concerned with the end-user
transition here. Prior to the introduction of DOM level 1 attributes in
htmlparser2 ([which occurred over 3 years
ago](https://github.com/fb55/domhandler/pull/4), believe it or not), Cheerio was
silent about the "node-like" objects that Cheerio collections contained. Since
then, [we've carefully documented only the standard
attributes](https://github.com/cheeriojs/cheerio/commit/62d03c427cde2b1aca8b94a742437dc0ed96a37d).
So in SemVer terms, removing the non-standard attributes would technically be
valid for even a patch release. Even I can see that would be a little harsh,
but it does make the idea of removing them in a major release seem fair enough.

So to sum up, here's what I'm proposing for our road map:

1. We update this patch to allow parsing configuration such that `htmlparser2`
   is *only* used for XML parsing
2. We land this as a backwards-breaking change to a dedicated `v1.0` branch,
3. We decide whether we want to switch to sax-js
4. We release version 1.0

Does that sound reasonable to you two?

inikulin:
> Does that sound reasonable to you two?

Works for me. @fb55, do you have any objections?

fb55:
Sounds reasonable, let's do it :)

paambaati:
Would this still let us use `htmlparser2` in case we need it? We use `cheerio` heavily to scrape websites and speed is more important to us than spec-compliance — if I understand correctly from these discussions ([[1](https://github.com/inikulin/parse5/issues/102)], [[2](https://github.com/inikulin/parse5/issues/89)]), `parse5` is a little slower than `htmlparser2`.

inikulin:
@paambaati Yes, it's a bit slower and it will always be. I see two ways how we can address this concern:
- Keep `useHtmlParser2` option
- Add forgiving parser on top of parse5 SAXParser, it should be close to performance of `htmlparser2`

inikulin:
Regarding https://github.com/inikulin/parse5/issues/102 slow down was caused by excessive memory use and significant GC pauses. Now, SAXParser just ~1.2-1.5 slower than htmlparser2

paambaati:
I agree, but it would be great if there was some option to retain `htmlparser2` for some of us that would still need it.

@inikulin Also, as an aside, great work on `parse5`. I've been closely watching the project for a while and it has matured incredibly well.

fb55:
@paambaati Cheerio will still be able to work on top of a htmlparser2 DOM (especially as long as that DOM is still used for XML), but you would need to parse documents in a separate step.

The main issue with supporting multiple parsers are the additional steps for error reports. Allowing advanced users to use htmlparser2 DOMs is not in conflict with this.

inikulin:
I've changed options format. Please, review current code changes and then we'll consider moving further. \cc @fb55 @jugglinmike

fb55:
This is still using the plain htmlparser2 DOM, which is not ideal, otherwise lgtm. Let's move that to a separate step to move things ahead here.

We could add some validation for the options as a couple of them won't work anymore, although having some notice about upgrading to 1.0 should be sufficient.

inikulin:
@fb55 
> This is still using the plain htmlparser2 DOM, which is not ideal, otherwise lgtm. Let's move that to a separate step to move things ahead here.

As far as I remember we've decided to postpone this change. Since we still use `htmlparser2` for XML I don't see how it's even possible at this point. However, once we switch to SAX there should  be no problem with that task.

> We could add some validation for the options as a couple of them won't work anymore, although having some notice about upgrading to 1.0 should be sufficient.

Actually, current change is back compatible. Old configuration still works. Although we can add deprecation warnings if old API is used. Regarding notice about upgrade: where should we place it? In Readme?

fb55:
@inikulin `domhandler` has an option for adding DOM v1-compatible properties to the DOM, which is turned on by default inside of cheerio.

What do you think is the better option: Add parse5's tree builder API to `domhandler`, or provide a handler for parse5 itself?

inikulin:
@fb55 

> What do you think is the better option: Add parse5's tree builder API to domhandler, or provide a handler for parse5 itself?

I'm afraid I don't quite understand the question. Can elaborate a bit, please?

inikulin:
Ping @fb55

fb55:
@inikulin For some reason I never checked whether parse5 already supported the dom resulting from the `withDomLvl1` option, which it does. Sorry for holding this back!

fb55:
@jugglinmike Imho this is fine for the 1.0 release, let's push it out asap and add a note about potentially dropped properties to the readme. What do you think?

jugglinmike:
I'm excited about it, too, but I'd like to be a little more conservative. Would you be up to vetting 1.0 using a release candidate?

fb55:
A rc sounds good 👍

jugglinmike:
Great! I'm thinking it might make sense to manually merge this patch to a feature branch, just in case we need to release from `master` during the RC phase.

It will be a few days before I can help out with this. If you have the bandwidth to release yourself, then be my guest.

stevenvachon:
> switch to parse5 AST

Will this ever happen, like when/if we drop htmlparser2 for XML?

jugglinmike:
Alright, I've rebased this on the latest version of `master`, merged it to [a new branch named `v1.0.0`](https://github.com/cheeriojs/cheerio/tree/v1.0.0) and [released it as version `1.0.0-rc.1`](https://github.com/cheeriojs/cheerio/releases/tag/1.0.0-rc.1). Let's see what our users think!

inikulin:
@jugglinmike What about docs? E.g. `$(html)` and `$.load(html)` now have different behavior: first one parses markup as the fragment, whereas former parses markup as a document.

vvo:
@jugglinmike Nice! Did you do any benchmarks? Just being curious here.

jugglinmike:
@inikulin I assumed the patch's documentation changes were complete, but you're right--they are not. If the 2-week delay following my original commitment is any indication, I'm stretched a little thin at the moment. Do you have the bandwidth to address this?

inikulin:
@jugglinmike 
> Do you have the bandwidth to address this?

Unfortunately, no. I'm dead busy at the moment as well. 😞

jugglinmike:
Alright, a draft document is available for review at https://github.com/cheeriojs/cheerio/pull/1027.

jugglinmike:
@vvo We continue to maintain a benchmark suite within the `bench/` subdirectory of this project, but (as noted in gh-700), it is a little dated. I'd love to revive it and would gladly provide assistance to anyone interested in helping out. Might that be you?

jugglinmike:
> > We could add some validation for the options as a couple of them won't work
> > anymore, although having some notice about upgrading to 1.0 should be
> > sufficient.
>
> Actually, current change is back compatible. Old configuration still works.
> Although we can add deprecation warnings if old API is used. Regarding notice
> about upgrade: where should we place it? In Readme?

@inikulin It looks like @fb55 was correct. See the following issue comment:

https://github.com/cheeriojs/cheerio/issues/1029#issuecomment-307589537

In that case, if the internal `useHtmlParser2` option is omitted, the
`decodeEntities` option is not honored. Likewise, `normalizeWhitespace` does
not have an effect without `useHtmlParser2`. Browsing through the Parse5 source
code, I can't find any references to these options.

inikulin:
OK, I'm a bit confused at this point. parse5 doesn't have this options and I don't have any plans supporting them, since they are not spec-conforming. The first implementation of this PR used a fallback to htmlparser2 when these options are specified. Then we decided to avoid such implicit switching of parsers. Any ideas how we can handle it now to avoid confusion?

stevenvachon:
@jugglinmike It's my opinion that `normalizeWhitespace`does not belong in cheerio. Cheerio is for parsing and traversing/scraping. Optimizations should be done with something like [html-minifier](https://npmjs.com/html-minifier) or [htmlnano](https://npmjs.com/htmlnano).

jugglinmike:
> The first implementation of this PR used a fallback to htmlparser2 when these
> options are specified. Then we decided to avoid such implicit switching of
> parsers. Any ideas how we can handle it now to avoid confusion?

Now that we're disallowing `htmlparser2` use generally, `xmlMode` is
superfluous. Parsing options *only* apply when parsing as XML, so I think the
best way to avoid confusion is to combine the `xml` and `xmlMode` options. We
can support all the use cases by simply exposing an option named `xml` that is
used as follows:

`xml` value | Effect
------------|-------
`undefined` | parse input as HTML using parse5
`true`      | parse input as XML using htmlparser2 with the default parsing options
Object      | parse input as XML using htmlparser2 with the options specified by the object

I'm hoping this will avoid the kind of confusion voice [in this
discussion](https://github.com/cheeriojs/cheerio/pull/1033/files#r123777233):

> Then I am really confused because the README says the default parser is
> parse5 then we say the options are htmlparser2, hard to make sense of this.

Internally, we'll still merge the parsing options to the top-level object and
set `xmlMode: true` (in order to satisfy htmlparser2), but that is an
implementation detail that we don't need to document.

I've pushed a commit to https://github.com/cheeriojs/cheerio/pull/1033 that
implements this change and updates the documentation accordingly. Please let me
know what you think!

gajus:
Is there a maintained fork with this PR?

fb55:
I'm merging this into master to resolve all of the linked issues. Afterwards, I will remove the master branch, rename the current `1.0.0` branch to `main` and have that be the new top branch.

These changes have already been merged into `1.0.0`, so this PR resolved. Thanks a lot @inikulin for putting it up in the first place!

dor-benatia:
@inikulin can you please explain how this removed the recursive node decoration calls ?

fb55:
@dor-benatia This is an independent change; see #1559.

PR #999, opened 2017-03-31
Update .filter function docs.

konstantin-popov:
Callback passed to function `filter` get called with 2 arguments (`index` and `element`).


fb55:
Thanks @konstantin-popov!

PR #1002, opened 2017-04-01
Standardise readme on ES6 variable declarations (closes #1001)

### LINKED ISSUE #1001: Docs: Readme uses a mix of ES5 and ES6 variable declarations
The readme uses a mix of ES5 and ES6 variable declarations (var & let) I think this should be standardised on ES6 so that the docs are consistent and that the let should be replaced by const where possible as this seems to be the generally accepted best practice.

[JavaScript ES6+: var, let, or const?](https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75)
[How much should I be using 'let' vs 'const' in ES6?](http://softwareengineering.stackexchange.com/questions/278652/how-much-should-i-be-using-let-vs-const-in-es6)
[ES6 const is not about immutability](https://mathiasbynens.be/notes/es6-const)

Pull request to follow

dekatron:
- Previously docs contained examples using a mix of ES5 and ES6 syntax as well as some missing variable declarations.
- The first example (that did use ES6) used let instead of const, it seems to be considered best practice to use const when ever possible (i.e. when a pointer is not going to change, even if it's value does in the case of objects an arrays)

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -23,8 +23,8 @@
 <br />
 
 ```js
-let cheerio = require('cheerio')
-let $ = cheerio.load('<h2 class="title">Hello world</h2>')
+const cheerio = require('cheerio')
+const $ = cheerio.load('<h2 class="title">Hello world</h2>')
 
 $('h2.title').text('Hello there!')
 $('h2').addClass('welcome')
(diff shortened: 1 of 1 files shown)


fb55:
Thanks @dekatron!

dekatron:
No worries

PR #1015, opened 2017-04-30
Clarify project scope in README file

jugglinmike:
The project frequently receives bug reports from new users who expect
full web browser behavior. The project documentation does not make it
particularly clear that such functionality is out-of-scope for the
project. It includes some text that is specific to the JSDom project,
but that explanation assumes familiarity with JSDom, and its location
at the end of the file makes it easy to miss.

Explicitly describe the scope of the project in the introductory section
of the README file, prior to API documentation and remove the
JSDom-specific text.


fb55:
The jsdom section was probably outdated, and this is a good clarification for new users.

jugglinmike:
Glad you agree :) Thanks for the review and the merge!

PR #1016, opened 2017-05-03
Test suite housekeeping

DianeLooney:
Added a few tests to increase coverage by a small amount.
Cleaned up attributes.val() tests to match the existing style.
Removed a small unreachable if statement.


fb55:
Awesome changes @DianeLooney, thanks a lot!

Issue #986, opened 2017-02-22
.text() includes text within <script> and <style>


Vittoriourzi:
cheerio .text() not strip inline style and scripts.
If i perform a $("html").text() for strips all tags, the script include in text the textual content inside the inline script.

example of an html.

```html
<html><head></head><body>
<div>ciao</div>
<script>function(){blablabla}</script>
</body></html>
```

if a do: $("html").text() the ouput will be:
ciao function(){blablabla}

But in jQuery the inline tags were strip and removed too!

henrix343:
@Vittoriourzi does cherio suport innerText ?

matthewmueller:
This is currently implemented, but may be reverted. See: https://github.com/cheeriojs/cheerio/issues/1050

PR #1018, opened 2017-05-05
.text() ignores script and style tags

Dreamiko:
[ First time pull :) ]

text function will ignore script and style tags

Should solve this problem: https://github.com/cheeriojs/cheerio/issues/986


fb55:
Definitely a needed change, thanks @Dreamiko!

JimmyRobz:
Just updated cheerio to 1.0.0-rc.1, this change broke one of my apps. Fyi, jQuery does return text inside script tags.

PR #1003, opened 2017-04-11
Add nested wrapping test

whodidthis:
Not sure if this is correct but I think cheerio is still missing the nested wrapping feature from jQuery:

```
<html>
<body>
<img src="image.gif">
</body>
</html>

$('img').wrap('<div class="figure-wrapper"><figure></figure></div>');
=>

<html>
<body>
<div class="figure-wrapper">
  <figure>
    <img src="image.gif">
  </figure>
</div>
</body>
</html>

PR #1019, opened 2017-05-08
Added inner wrap support

DianeLooney:
Added the test-cases given in [PR#1003](https://github.com/cheeriojs/cheerio/pull/1003).
Added support for nested wrapping.
Current functionality is to follow the path of first children in the wrapper, and place the wrapped node there.

I screwed up my branching so this commit will look a bit strange.


jugglinmike:
I cleaned up the history a bit (removing unrelated commits and formally introducing @whodidthis's contribution) and followed up with some code style corrections. It's all merged to `master` at commit 124a8b89eacb5179b59c3c84b9b1804e8ed29864. Thanks for the patch!

PR #1027, opened 2017-06-03
Document release notes with new `HISTORY.md` file

jugglinmike:
@fb55 @inikulin What do you think? Did I miss anything? Would a separate `MIGRATING.md` file be more clear?

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,0 +1,43 @@
+# Cheerio releae notes
+
+## 1.0.0-rc.1 (2017-05-27)
+
+This release changes Cheerio's default parser to [the Parse5 HTML
+parser](https://github.com/inikulin/parse5). Parse5 is an excellent project
+that rigorously conforms to the HTML standard. It does not support XML, so
+Cheerio continues to use [htmlparser2](https://github.com/fb55/htmlparser2/)
+when working with XML documents.
+
+This switch addresses many long-standing bugs in Cheerio, but some users may
+experience slower behavior in performance-critical applications. In addition,
+htmlparser2 is more forgiving of invalid markup which can be useful when input
+sourced from a third party and cannot be corrected. For these reasons, this
+release offers users the ability to continue using htmlparser2 through explicit
+configuration.
+
+### Migrating from 0.x
+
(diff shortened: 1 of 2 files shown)


jugglinmike:
Woa. Thanks @coveralls

HISTORY.md
```diff
@@ -0,0 +1,43 @@
...
+This switch addresses many long-standing bugs in Cheerio, but some users may
+experience slower behavior in performance-critical applications. In addition,
+htmlparser2 is more forgiving of invalid markup which can be useful when input
+sourced from a third party and cannot be corrected. For these reasons, this
+release offers users the ability to continue using htmlparser2 through explicit
+configuration.
+
+### Migrating from 0.x
+
+**Re-enabling htmlparser2** As noted above, the new Parse5 library may not be
```
inikulin: Actually we've changed API a bit according to https://github.com/cheeriojs/cheerio/pull/985#issuecomment-289225391

Officially there is only one option: `xml` that can be either boolean or an object that contains htmlparser2 settings. However, for back compatibility, I've left ability to provide options on the `options`. `useHtmlParser2` option is used internally for DOM cloning, and we decided to not make it a part of public, however I don't have any problems with that.

inikulin: Actually we've changed API a bit according to https://github.com/cheeriojs/cheerio/pull/985#issuecomment-289225391

Officially there is only one option: `xml` that can be either boolean or an object that contains htmlparser2 settings. However, for back compatibility, I've left ability to provide options on the `options`. `useHtmlParser2` option is used internally for DOM cloning, and we decided to not make it a part of public, however I don't have any problems with that.

jugglinmike: Thanks for the review; it looks like I was a bit sloppy, likely due to my
limited availability. I was also confused because the readme currently states,

> If any of these options is set to non-default value cheerio will implicitly
> use `htmlparser2` as an underlying parser. In addition, you can use
> `useHtmlParser2` option to force cheerio use `htmlparser2` instead of
> `parse5`.

I'm going to remove this change.

And it turns out I was negligent in more ways than one: I completely forgot
that this project *already* maintains release notes. I'm opening a new pull
request to modify that file instead, incorporating your suggestions, of course.

Issue #1031, opened 2017-06-05
Automatically adding <html><head><body> ?


sandro-pasquali:
OSX 10.12.4
Node 8.0
"cheerio": "1.0.0-rc.1"

Not sure if this is expected, but:

```
let $ = cheerio.load('<div>Hello</div>');
console.log($.html());

// <html><head></head><body><div>Hello</div></body></html>
```
While in my case the auto-wrapping as an `<html>` document is weirdly useful, I was still surprised to see it. Is this a bug, or a feature? I don't see a `.fragment()` method or anything like that. Maybe just missed something.

Thanks!

jugglinmike:
Yup, this is a breaking change for version 1.0. We're working on documentation here: https://github.com/cheeriojs/cheerio/pull/1027

sandro-pasquali:
Oh, thanks. 

Re: performance here: <https://github.com/cheeriojs/cheerio/commit/6aae08413e997ec06448b744643f6bc00a00f4b9>

> This switch addresses many long-standing bugs in Cheerio, but some users may
>  experience slower behavior in performance-critical applications.

Just so I understand, is this saying that setting `useHtmlParser2` will often result in better performance than not doing so? 

And more specifically, if my operation is **always** modifying a small fragment (something like `$('<div>hi</div>').append(...)`) which is the best parser to choose in terms of performance?

Thanks for the great library.

jugglinmike:
My apologies; the `useHtmlParser2` option is not considered "public" API and
should not be relied upon. The latest version of the documentation (currently
under review at gh-1033) includes an updated recommendation for using
`htmlparser2` with Cheerio.

When it comes to performance in any specific scenario, the best way to verify
is to run some trials locally.

xymbol:
>My apologies; the useHtmlParser2 option is not considered "public" API and should not be relied upon.

Would be great to add more tests extracted from valid use cases to make sure users don't end up locked with < 1.0. The number of changed tests suggest that a lot of apps will break with 1.0.

Just posted one example test here. https://github.com/cheeriojs/cheerio/issues/1029#issuecomment-307589537

fidian:
I also deal with fragments, but I could just as easily work with full documents. If it was consistently fragments then I could just grab the elements in `<body>`. If it was consistently full documents then there's potentially no change, though I may need to remove a `<head>` or `<body>` element in order to get back to what was supplied.

[...]

PR #1033, opened 2017-06-10
V1 release notes 2

jugglinmike:
@fb55 @inikulin here's a second attempt, hopefully more in-line with reality. Note the changes to the "pre-release" script; I thought this was the best way to promote safety while allowing us to make manual modifications to the project's release history document.

Also, because the lack of documentation has already caused confusion for early adopters, I'm recommending we publish this as a new release candidate. Following that, I'd like to update the `Readme.md` file in `master` to let folks know that we're currently vetting a release candidate, and pointing them to the `v1.0.0` branch.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -908,7 +908,7 @@ Checks to see if the `contained` DOM element is a descendant of the `container` 
 Parses a string into an array of DOM nodes. The `context` argument has no meaning for Cheerio, but it is maintained for API compatability.
 
 #### $.load( html[, options ] )
-Create a jQuery-like function, bound to a document created from the provided markup. Note that similar to web browser contexts, this operation may introduce `<html>`, `<head>`, and `<body>` elements. See the previous section titled "Loading" for usage information.
+Create a querying function, bound to a document created from the provided markup. Note that similar to web browser contexts, this operation may introduce `<html>`, `<head>`, and `<body>` elements. See the previous section titled "Loading" for usage information.
 
 ### Plugins
 
--- a/Readme.md
+++ b/Readme.md
@@ -121,6 +121,23 @@ These parsing options are taken directly from [htmlparser2](https://github.com/f
 For a full list of options and their effects, see [this](https://github.com/fb55/DomHandler) and
 [htmlparser2's options](https://github.com/fb55/htmlparser2/wiki/Parser-options).
 
+Some users may wish to parse markup with the `htmlparser2` library, and
+traverse/manipulate the resulting structure with Cheerio. This may be the case
+for those upgrading from pre-1.0 releases of Cheerio (which relied on
+`htmlparser2`), for those dealing with invalid markup (because `htmlparser2` is
+more forgiving), or for those operating in performance-critical situations
+(because `htmlparser2` may be faster in some cases).
+
+To support these cases, `load` also accepts a `htmlparser2`-compatible data
+structure as its first argument. Users may install `htmlparser2`, use it to
+parse input, and pass the result to `load`:
+
+```js
+const htmlParser2 = require('htmlparser2');
+const dom = { /* See htmlParser2's documentation for usage instructions. */ };
+const $ = cheerio.load(dom);
+```
+
 ### Selectors
 
 Cheerio's selector implementation is nearly identical to jQuery's, so the API is very similar.
--- a/Readme.md
+++ b/Readme.md
@@ -127,8 +127,9 @@ for those upgrading from pre-1.0 releases of Cheerio (which relied on
 `htmlparser2`), for those dealing with invalid markup (because `htmlparser2` is
 more forgiving), or for those operating in performance-critical situations
 (because `htmlparser2` may be faster in some cases). Note that "more forgiving"
-is not necessarily an advantage; `htmlparser2` will recover from some invalid
-markup differently than web browsers, and this may or may not be helpful.
+means `htmlparser2` has error-correcting mechanisms that aren't always a match
+for the standards observed by web browsers. This behavior may be useful when
+parsing non-HTML content.
 
 To support these cases, `load` also accepts a `htmlparser2`-compatible data
 structure as its first argument. Users may install `htmlparser2`, use it to
@@ -136,7 +137,7 @@ parse input, and pass the result to `load`:
 
 ```js
 // Usage as of htmlparser2 version 3:
-const htmlParser2 = require('htmlparser2');
+const htmlparser2 = require('htmlparser2');
 const dom = htmlparser2.parseDOM(document, options);
 
 const $ = cheerio.load(dom);
--- a/Readme.md
+++ b/Readme.md
@@ -126,15 +126,19 @@ traverse/manipulate the resulting structure with Cheerio. This may be the case
 for those upgrading from pre-1.0 releases of Cheerio (which relied on
 `htmlparser2`), for those dealing with invalid markup (because `htmlparser2` is
 more forgiving), or for those operating in performance-critical situations
-(because `htmlparser2` may be faster in some cases).
+(because `htmlparser2` may be faster in some cases). Note that "more forgiving"
+is not necessarily an advantage; `htmlparser2` will recover from some invalid
+markup differently than web browsers, and this may or may not be helpful.
 
 To support these cases, `load` also accepts a `htmlparser2`-compatible data
 structure as its first argument. Users may install `htmlparser2`, use it to
 parse input, and pass the result to `load`:
 
 ```js
+// Usage as of htmlparser2 version 3:
 const htmlParser2 = require('htmlparser2');
-const dom = { /* See htmlParser2's documentation for usage instructions. */ };
+const dom = htmlparser2.parseDOM(document, options);
+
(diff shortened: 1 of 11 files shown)


Readme.md
```diff
@@ -106,22 +106,38 @@ const $ = cheerio.load('<ul id="fruits">...</ul>', {
...
+more forgiving), or for those operating in performance-critical situations
+(because `htmlparser2` may be faster in some cases).
+
+To support these cases, `load` also accepts a `htmlparser2`-compatible data
+structure as its first argument. Users may install `htmlparser2`, use it to
+parse input, and pass the result to `load`:
+
+```js
+const htmlParser2 = require('htmlparser2');
+const dom = { /* See htmlParser2's documentation for usage instructions. */ };
```
vvo: I believe we might want to dig a bit more here so I don't have to go through multiple documentation and try to find the right line to copy paste. Can we give more details here?

Readme.md
```diff
@@ -106,22 +106,38 @@ const $ = cheerio.load('<ul id="fruits">...</ul>', {
...
 
 ```
 
 For a full list of options and their effects, see [this](https://github.com/fb55/DomHandler) and
 [htmlparser2's options](https://github.com/fb55/htmlparser2/wiki/Parser-options).
 
+Some users may wish to parse markup with the `htmlparser2` library, and
+traverse/manipulate the resulting structure with Cheerio. This may be the case
+for those upgrading from pre-1.0 releases of Cheerio (which relied on
+`htmlparser2`), for those dealing with invalid markup (because `htmlparser2` is
```
vvo: "invalid markup" => A lot of websites may have invalid markup and cheerio is used in many situations for scraping it seems.

Does it means cheerio will fail more often? Is there a quick reference to understand what was "working" before that may fail now?

Readme.md
```diff
@@ -106,22 +106,38 @@ const $ = cheerio.load('<ul id="fruits">...</ul>', {
 });
 ```
 
-These parsing options are taken directly from [htmlparser2](https://github.com/fb55/htmlparser2/wiki/Parser-options), therefore any options that can be used in `htmlparser2` are valid in cheerio as well. If any of these options is set to non-default value cheerio will implicitly use `htmlparser2` as an underlying parser. In addition, you can use `useHtmlParser2` option to force cheerio use `htmlparser2` instead of `parse5`. The default options are:
+These parsing options are taken directly from [htmlparser2](https://github.com/fb55/htmlparser2/wiki/Parser-options), therefore any options that can be used in `htmlparser2` are valid in cheerio as well. The default options are:
```
vvo: Shouldn't it says that the parsing options are taken directly from parse5? Since the changelog says the default parser is parse5

inikulin: Nope, these are indeed `htmlparser2` options.

vvo: Then I am really confused :D

Readme.md
jugglinmike: This was intentional. I want to avoid unnecessary coupling between this document and the htmlparser2 project.

Readme.md
jugglinmike: Parse5 follows a standard, while htmlparser2 implements its own custom
heuristics. Because of this, the definition of what was "working" before is not
well understood.

Readme.md
jugglinmike: I've tried to address this by removing `xmlMode` and only documenting `xml`. @vvo can you let me know if this latest version of the readme is easier to understand?

jugglinmike:
Okay, I've pushed a few more commits up that I hope will address folks' recent
concerns:

- Extending the migration guide with more specific instruction for working with
  document fragments (hopefull help avoid the confusion discussed here:
  https://github.com/cheeriojs/cheerio/issues/1031)
- Consolidate the `xml` and `xmlMode` options (see [this
  comment](https://github.com/cheeriojs/cheerio/pull/985#issuecomment-310852805)
  for rationale)
- Rename `useHtmlParser2` option to help consumers avoid relying on library
  internals

I'd very much appreciate review from maintainers and consumers, but all of this
is continuing to cause confusion. I would like to merge and publish a new
release candidate no later than tomorrow. Sorry for the urgency; it seems
justified in this case, though.

jugglinmike:
It would certainly be nice to see that kind of speed from Parse5, but really, the main motivation for htmlparser2 is support for XML.

Thanks for looking this over!

jugglinmike:
Hey @fb55 would you mind giving this a look?

Readme.md
fb55: Perhaps add that browsers won't understand this either.

Readme.md
fb55: I agree with @vvo. `htmlparser2` won't change its API anymore — right now I consider it done. Changing this to `htmlparser2.parseDOM(document, options)` probably simplifies usage a lot here.

Readme.md
```diff
@@ -888,7 +904,7 @@ Checks to see if the `contained` DOM element is a descendant of the `container`
 Parses a string into an array of DOM nodes. The `context` argument has no meaning for Cheerio, but it is maintained for API compatability.
 
 #### $.load( html[, options ] )
-Load in the HTML. (See the previous section titled "Loading" for more information.)
+Create a jQuery-like function, bound to a document created from the provided markup. Note that similar to web browser contexts, this operation may introduce `<html>`, `<head>`, and `<body>` elements. See the previous section titled "Loading" for usage information.
```
fb55: Perhaps change _a jQuery-like function_ to something like _a querying function_? Otherwise this might seem off-putting for people that are not familiar with jQuery.

test/api/traversing.js
```diff
@@ -52,8 +52,8 @@ describe('$(...)', function() {
       expect(q('foo').find('> bar')).to.have.length(1);
     });
 
-    it('should query case-sensitively when in xmlMode', function() {
-      var q = cheerio.load('<caseSenSitive allTheWay>', {xmlMode: true});
+    it('should query case-sensitively when in xml', function() {
```
fb55: _xml mode_?

Readme.md
jugglinmike: Sounds good to me.

test/api/traversing.js
jugglinmike: Sure

Readme.md
jugglinmike: Well, if anyone is the authority on this, it's you! I'll make one minor precaution, just to be safe:

    // Usage as of htmlparser2 version 3:

Readme.md
jugglinmike: I agree that's useful to point out, but it's tricky to explain succinctly and in a way that will make sense to the average user. I've included a first attempt in a new commit:

> Note that "more forgiving" is not necessarily an advantage; `htmlparser2`
> will recover from some invalid markup differently than web browsers, and this
> may or may not be helpful.

jugglinmike:
Thanks for the review, @fb55! What do you think of this latest version?

Readme.md
fb55: I think the bigger point here is that htmlparser2 has error correction mechanisms that aren't always a match for the behaviour in browsers. parse5 on the other hand implements spec-compliant error correction.

The only people that will ever want to use htmlparser2 this way are the ones that want to parse non-html content. Not sure if we should even try to support that use-case.

I'm fine with keeping it like this for now and revising it if the number of related issues explodes.

Readme.md
```diff
@@ -96,32 +96,52 @@ const $ = require('cheerio');
...
+is not necessarily an advantage; `htmlparser2` will recover from some invalid
+markup differently than web browsers, and this may or may not be helpful.
+
+To support these cases, `load` also accepts a `htmlparser2`-compatible data
+structure as its first argument. Users may install `htmlparser2`, use it to
+parse input, and pass the result to `load`:
+
+```js
+// Usage as of htmlparser2 version 3:
+const htmlParser2 = require('htmlparser2');
```
fb55: Nit: Just saw that this line has an upper-case _p_

fb55:
One nit, otherwise looks great. Thanks a lot for doing this Mike!

Readme.md
jugglinmike: Good catch!

Readme.md
jugglinmike: I'm happy to keep iterating since it definitely still feels productive:

> Note that "more forgiving" means `htmlparser2` has error-correcting mechanisms that aren't always a match for the standards observed by web browsers. This behavior may be useful when parsing non-HTML content.

jugglinmike:
> One nit, otherwise looks great. Thanks a lot for doing this Mike!

Sure thing. I certainly appreciate your review :)

Readme.md
fb55: I like it 👍

jugglinmike:
Rebased and merged to the `v1.0.0` branch: https://github.com/cheeriojs/cheerio/releases/tag/1.0.0-rc.2

Thanks, all!

Readme.md
vvo: I was worried by parse5 failing more often than browsers but in the end if parse5 just works as my chrome browser, I am more than fine. Thanks

Issue #1077, opened 2017-09-09
Proposal for breaking change: removing the `root` element


jugglinmike:
### Background

Cheerio's `load` method accepts a string of markup and creates a selector
function. This function is "bound" to a document whose contents is a node
structure based on the input markup. This function is intended to behave much
like the global `jQuery`/`$` function provided by the jQuery library (as if the
library had been loaded in a document generated from the same markup by a web
browser).

Historically, Cheerio has always attached the parsed markup to a non-standard
"root" element (i.e. `<root>`). As far as I know, this was implemented to
support the `load` method's behavior when given markup describing a document
fragment--strings like `'<p>1</p><p>2</p>'` could be passed to `load` and still
produce a single top-level element.

Thanks to @inikulin and his Parse5 library, the release candidate for version
1.0.0 normalizes the parsing behavior of `load`. It *always* produces a
complete document--just a like a web browser. The result is much more
predictable, standards-compliant, and "familiar" to web developers. It's a
backwards-breaking change, though, and we hope to ease the upgrade path for
consumers through a concise migration guide.

### Proposal

Since we are already committed to a breaking change for version 1.0, I wanted
to consider making Cheerio's behavior even more browser-like. I'd like to get
rid of the `<root>` element and instead rely on the `<html>` element (which
again, is either described by the consumer's input markup or automatically
created by Parse5).

This would involve removing the Cheerio-specific `$.root()` method from the
API.  Users who previously used it as a basis for traversal could re-write code
like `$.root().find('div')` with `$('html').find('div')`.

However, many use cases involve rendering full documents. For this,
`$('html').html()` is not equivalent because the resultant string does not
include the document element itself. So we'll need to keep another
Cheerio-specific method: `$.html()`.

Other than that, I think it would just be matter of updating Cheerio's
internals to operate without `<root>`. I spent just enough time trying this out
to see that it is not trivial, but I don't believe that there is any technical
reason it can't be done.

I'd love to get feedback from any Cheerio user, but I'm particularly interested
in hearing from @fb55, @matthewmueller, and @inikulin. Do you think this is a
good idea? Do you think it would invalidate any existing use cases? Or is there
any other reason it isn't technically possible? Or more subjectively, do you
think the change would be too jarring for consumers to justify the benefit?

matthewmueller:
[...]

PR #1078, opened 2017-09-09
Re-write migration guide for version 1.0

jugglinmike:
Incorporate recent feedback from consumers who have experimented with
the version 1.0 release candidate.

---

@fb55 @ljharb: I've re-written the guide based on Jordan's experience migrating
the Enzyme project to the release candidate. Think this is an improvement? Is
there anything else you'd like to see here?

While writing this, I began to question the importance of Cheerio's `<root>`
element. I'm proposing a more drastic change for version 1.0 over at gh-1077.
I'd love to hear your thoughts on that, too!

### DIFF
--- a/History.md
+++ b/History.md
@@ -17,26 +17,51 @@ pattern.
 
 ### Migrating from version 0.x
 
-`cheerio.load( html[, options ] )`` This method continues to produce
-jQuery-like functions, bound to the provided input text. In prior releases, the
-provided string was interpreted as a document fragment. This meant that in a
-statement such as:
+`cheerio.load( html[, options ] )` This method continues to act as a "factory"
+function. It produces functions that define an API that is similar to the
+global `jQuery` function provided by the jQuery library. The generated function
+operates on a DOM structure based on the provided HTML.
+
+In releases prior to version 1.0, the provided HTML was interpreted as a
+document fragment. Following version 1.0, strings provided to the `load` method
+are interpreted as documents. The same example will produce a `$` function that
+operates on a full HTML document, including an `<html>` document element with
+nested `<head>` and `<body>` tags. This mimics web browser behavior much more
+closely, but may require alterations to existing code.
+
+For example, the following code will produce different results between 0.x and
+1.0 releases:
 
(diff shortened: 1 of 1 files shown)


History.md
```diff
@@ -17,26 +17,51 @@ pattern.
...
-interpreted as documents. The same example will produce a `$` function that
-operates on a full HTML document, including an `<html>` document element with
-nested `<head>` and `<body>` tags. This mimics web browser behavior much more
-closely, but may require alterations to existing code. To work with fragments,
-first load an empty document, and then use the resulting `$` to parse the
-input:
+    //=> In version 0.x: '<p>Hello, <b>world</b>!</p>'
+    //=> In version 1.0: '<html><head></head><body><p>Hello, <b>world</b>!</p></body></html>'
 
+Users wishing to parse, manipulate, and render full documents should not need
```
ljharb: This is much clearer - in other words, the breaking change isn't `.load()`, it's the *combination* of `.load()` and `.root()`.

jugglinmike:
> Does that mean that `cheerio.load(html).root().anyDescendantCheerioFunctionHere(…)` can be replaced with `cheerio.load('')('<div />').append(html).anyDescendantCheerioFunctionHere(…)`, but `cheerio.load(html).root()` still has no true replacement?

In `cheerio.load(html).root()`, if `html` is a string of markup describing a complete document, then `cheerio.load(html).root()` should behave identically in version 1.0.

If `html` is markup for a document fragment, then it's not so easy. That's because jQuery does not provide an API for creating document fragments, and Cheerio Cheerio *mostly* doesn't. I've been thinking about this a lot these days, and I've just now identified a workaround: you *could* re-set the contents of the `<root>` element after it has been created with a complete document via `load`. This would be equivalent: `cheerio.load('').root().html(html)`. But I only offer that in the interest of transparency; [in the issue I referenced above, I'm actively proposing the outright removal of the `<root>` element](https://github.com/cheeriojs/cheerio/issues/1077) precisely because you can't do this in jQuery. So I'd hold back on using that approach for now.

jugglinmike:
Also thanks for the review!

PR #1079, opened 2017-09-15
Dom serializer bump

gauravmuk:


fb55:
Looks like this was solved already

PR #1111, opened 2017-11-21
Add Missing serialize Documentation

dhurlburtusa:
I noticed that the `serialize` method was not documented. I didn't see anything in the code indicating that it was deprecated and only exists for backwards compatibility.


fb55:
Thanks a lot!

PR #1135, opened 2018-01-27
Prevent data-uri splitting at css parsing

carrieje:
This fix prevent `url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAC4AAA)`
properties to be split in two properties, due to the semi-colon it contains.

It does not introduce any regression.
It may slow a bit the process as it relies on Regexp now.

It does not handle every data-uri types.
As of RFC2397[1] and RFC2396[2], other fields such as `mediatype` may
contain semi-colon.

It is thus still not perfect, but fix at least some cases.

[1] https://tools.ietf.org/html/rfc2397
[2] https://tools.ietf.org/html/rfc2396

### DIFF
--- a/lib/api/css.js
+++ b/lib/api/css.js
@@ -110,7 +110,7 @@ function parse(styles) {
   if (!styles) return {};
 
   return styles
-    .split(';')
+    .split(/;(?!base64)/)
     .reduce(function(obj, str){
       var n = str.indexOf(':');
       // skip if there is no :, or if it is the first/last character
--- a/test/api/css.js
+++ b/test/api/css.js
@@ -32,6 +32,11 @@ describe('$(...)', function() {
       expect(el.css('background-image')).to.equal('url(http://example.com/img.png)');
     });
 
+    it('(prop): should not mangle embedded data-uris', function() {
+      var el = cheerio('<li style="background-image:url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAC4AAA);">');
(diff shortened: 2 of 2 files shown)


fb55:
I would prefer a more generalised solution. Is there some css parser we can use for this?

DianeLooney:
the npm module 'css' seems to do the trick at first glance:

```js
let css = require('css');
var obj = css.parse('body { font-size: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAC4AAA); font-weight: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAC4AAA);}');
console.log(obj.stylesheet.rules[0])
```
```
{ type: 'rule',
  selectors: [ 'body' ],
  declarations:
   [ { type: 'declaration',
       property: 'font-size',
       value: 'url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAC4AAA)',
       position: [Object] },
     { type: 'declaration',
       property: 'font-weight',
       value: 'url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAC4AAA)',
       position: [Object] } ],
  position:
   Position {
     start: { line: 1, column: 1 },
     end: { line: 1, column: 150 },
     source: undefined } }
```
I'll look into getting the css processor changed then open a pr for it. I'm guessing this change will be somewhat involved.

fb55:
That would be amazing — lmk if there is something to review :)

fb55:
Closing this due to lack of activity. Happy to revisit the topic should someone volunteer.

PR #1139, opened 2018-02-06
Fix data attribute retrieving problem.

tai2:
Where original markup has a data attribute 'x' and it is overriden
by data('x', 'a'), it is accidentaly reset to the original markup state once
data() is called to get all data attributes. This commit fix the problem.


fb55:
Thanks a lot @tai2!

PR #1141, opened 2018-02-07
baseURI

bluelovers:
```ts
import * as cheerio from '..';

let html = `<body><a href="/n2627t/1/"/></body>`;

let $1 = cheerio.load(html);
let $2 = cheerio.load(html, {
	baseURI: 'http://ncode.syosetu.com',
});

console.log($1('a').prop('href')); // => /n2627t/1/
console.log($2('a').prop('href')); // => http://ncode.syosetu.com/n2627t/1/
```

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -126,7 +126,7 @@ exports.prop = function (name, value) {
         break;
       case 'href':
       case 'src':
-        let allowUrlResolve = {
+        var allowUrlResolve = {
           'href': ['A'],
           'src': ['IMG', 'IFRAME'],
         };
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -127,6 +127,7 @@ exports.prop = function (name, value) {
         property = this[0].name.toUpperCase();
         break;
       case 'href':
+      case 'src':
         property = getProp(this[0], name);
         if (this.options.baseURI) {
           property = (new URL(property, this.options.baseURI)).href;
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -26,7 +26,7 @@ var $ = require('../static'),
   // Matches strings that look like JSON objects or arrays
   rbrace = /^(?:\{[\w\W]*\}|\[[\w\W]*\])$/;
 
-const URL = require('jsdom-url').URL;
+var URL = require('jsdom-url').URL;
 
 var getAttr = function(elem, name) {
   if (!elem || !isTag(elem)) return;
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -133,7 +133,7 @@ exports.prop = function (name, value) {
 
         property = getProp(this[0], name);
 
-        if (this.options.baseURI && allowUrlResolve[name] && allowUrlResolve[name].includes(this.prop('tagName'))) {
+        if (this.options.baseURI && allowUrlResolve[name] && allowUrlResolve[name].indexOf(this.prop('tagName')) != -1) {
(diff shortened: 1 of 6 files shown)


fb55:
Tests are failing

bluelovers:
hmm i can't do anymore for low version..............
hope someone can make a version for this

fb55:
Would be great if this 1) was optional and 2) had another test case. You probably also want to resolve `src` attributes.

bluelovers:
@fb55 this already is optional if we didn't set baseURI

fb55:
Okay, I see. So whatwg-url depends on node >= 6, while cheerio still supports node@0.12. Can we just use [`url.resolve`](https://nodejs.org/api/url.html#url_url_resolve_from_to) then? Also, would you mind adding a test case for this?

lib/api/attributes.js
```diff
@@ -125,6 +126,13 @@ exports.prop = function (name, value) {
       case 'nodeName':
         property = this[0].name.toUpperCase();
         break;
+      case 'href':
+      case 'src':
+        property = getProp(this[0], name);
```
fb55: Also, move this into the `if` block?

bluelovers: what mean for into if?
u mean move this code to case default and make a if block?

but i think is no need

the default is `property = getProp(this[0], name);`

and when we didn't set baseurl
also is  `property = getProp(this[0], name);`

bluelovers:
alright should be ok now

lib/api/attributes.js
```diff
@@ -125,6 +124,19 @@ exports.prop = function (name, value) {
       case 'nodeName':
         property = this[0].name.toUpperCase();
         break;
+      case 'href':
+      case 'src':
+        var allowUrlResolve = {
```
fb55: Can this be converted to

```js
(name === 'href' && tagName === 'a') ||
(name === 'href' && (tagName === 'img' || tagName === 'iframe'))
```

inside of the `if` clause?

lib/api/attributes.js
```diff
@@ -125,6 +124,19 @@ exports.prop = function (name, value) {
...
+          'href': ['A'],
+          'src': ['IMG', 'IFRAME'],
+        };
+
+        property = getProp(this[0], name);
+
+        if (this.options.baseURI && allowUrlResolve[name] && allowUrlResolve[name].indexOf(this.prop('tagName')) != -1) {
+          property = utils.urlResolve(property, this.options.baseURI);
```
fb55: Can we use `url.resolve` directly here?

lib/api/attributes.js
fb55: Also, this doesn't cover a bunch of cases, but imho that's okay for now.

mmontag:
This would be nice to have!

test/api/url.resolve.js
```diff
@@ -0,0 +1,34 @@
...
+describe('$(...)', function()
+{
+  var $;
+
+  describe('.prop', function ()
+  {
+    beforeEach(function ()
+    {
+      $ = cheerio.load(links, {
+        baseURI: 'http://ncode.syosetu.com/a/b/c',
```
fb55: This should be replaced with `example.com`

fb55:
Would be great if someone could pick this up. There are currently three blockers:

- Needs a rebase
- Unnecessary indirection, the `url.resolve` call can be inlined.
- Links point to a private website, should be replaced

bluelovers:
i don't know why u not manual fix when merge this pr

PR #1152, opened 2018-02-27
Use eslint & prettier, add precommit hook

fb55:


fb55:
Node 0.12 isn't supported by this toolchain. I know @matthewmueller would prefer to keep the code base ES5-compliant, everything else is moving towards modern JS though. I agree that cheerio should never be transpiled, we should be fine targeting actively supported node versions, though. 

@jugglinmike What do you think about this? I'd like to target node@6, which would allow us to do quite a bit of cleanup.

fb55:
ES5 is still enforced by eslint. A lot of the tooling doesn't support node@0.12 anymore, so continuing to test on it would mean we are stuck with old versions for a lot of packages.

jugglinmike:
I would have liked to stick with JSHint, but I'm a bit biased. No worries.

fb55:
Had hoped for some feedback on this — I like prettier's autoformatting approach and eslint's plugin system a lot in general, but also haven't kept up with jshint in more than two years.

All good as long as you are okay with this though :)

jugglinmike:
Totally. Those projects are much more fully-featured, so I don't hold it against you

matthewmueller:
@fb55 looks good and sorry for the delay! I recently switched [nightmare](https://github.com/segmentio/nightmare/commit/88cfd473af340e34dcce035addab60e5c84cf0dd) to almost exactly the same setup :-)

Just to clarify my position a bit, many times I've seen transpiling improve the experience for the developer writing the code at the expense of everyone who installs and contributes the code. I'm also fine with upgrading the codebase a bit, but it seems like we're giving up compatibility for conciseness.

These decisions are often at odds with each other, I mostly just want to make sure that the folks that use cheerio have a good experience and it's easy enough for us to maintain.

PR #1159, opened 2018-03-11
Integrate markdown-magic to auto-generate TOC and contributors

fb55:
Right now both of these need to be manually updated.


fb55:
@jugglinmike Any issues with this? It helps with keeping the readme up to date and allows us to get more creative with what we include. We could even embed the jsdoc in the readme through https://github.com/bradtaylorsf/markdown-magic-jsdoc, even though that might be a bit heavy.

jugglinmike:
My inclination is to go the other way and remove the contributor list. At that point, the readme would be manageable enough to not really need a table of contents. I'm willing to compromise on what is admittedly a more subjective issue, but I'm not comfortable incorporating a process that is not trivially repeatable. Our release cadence has slowed down to the point where I expect to forget any manual steps, so I'd much prefer to nail those details down in the Makefile.

Also, note that the `make` target doesn't seem to function as intended:

    $ make readme
    Starting markdown-magic README.md
    Error: No files matched
    No files matched pattern README.md

It's strange that the binary exits with 0 in this case. But in any event, it looks like we need to use the `--path` option since our readme file is named `Readme.md` and not `README.md`:

    $ ./node_modules/.bin/md-magic --help

      Usage: md-magic [options]

      Options:

        -V, --version        output the version number
        -p, --path [path]    Define path to markdown (single path or glob). Default README.md (default: README.md)
        -i, --ignore [path]  (Optional) Define path to ignore (default: ./markdown.config.js)
        -c, --config [path]  (Optional) Define config file path. If you have custom transforms or a callback you will want to specify this option (default: null)
        -h, --help           output usage information

When I run this locally, though, the utility creates a contributor list which differs significantly than what is currently defined on the branch. Because Ihad to modify the GNU Make target, it's possible that my invocation differs from yours in some other significant way. Could you verify the effect of the following command?

    ./node_modules/.bin/md-magic --path Readme.md

fb55:
I've update the command, only difference is that one contributor was added. Sorry for missing this, and weird that you get different output.

> My inclination is to go the other way and remove the contributor list.

Totally fine with that. Just wanted to add a way to automatically generate it as it hasn't been updated for quite some time.

We could also leave the section empty and generate it together with the documentation (which also includes the readme). Nothing would be shown in the repo, but we would still mention contributors on the website.

gajus:
Please consider https://github.com/gajus/gitdown.

fb55:
I removed the _Contributors_ section from the readme as an alternative to this PR.

PR #1161, opened 2018-03-11
Ensure text nodes expose the DOM level 1 API

fb55:
Rebased version of #588. We can still optimise performance later on, this works for now.

### DIFF
--- a/lib/api/manipulation.js
+++ b/lib/api/manipulation.js
@@ -412,7 +412,7 @@ exports.text = function(str) {
     // Function support
     var self = this;
     return domEach(this, function(i, el) {
-      return exports.text.call(self, str.call(el, i, $.text([el])));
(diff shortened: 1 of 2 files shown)


jugglinmike:
Something seems to be wrong here, since the tests are failing on TravisCI. I'm spent for the evening, so I can't be much more constructive. I'll try to take a look in the coming days.

fb55:
The issue is how options are read. I wanted to look into it tomorrow, hope you're having a great evening!

fb55:
We used to pass an array as the context to `.text` when a function was passed, which broke things as that array didn't have an `options` property. I fixed this by passing a cheerio object instead of the array.

jugglinmike:
Thanks! I've created gh-1166 so we don't forget about the performance issue or how we'd like to resolve it.

PR #1162, opened 2018-03-11
Do not lint files excluded from version control

jugglinmike:
This includes code coverage reports as generated by the command `make
test-cov`.


fb55:
Whee good catch!

PR #1164, opened 2018-03-13
Use parse5 to serialise the DOM, use lodash to clone dom

### LINKED ISSUE #1101: find() convert `xlink:href` to `href`
`cheerio@^1.0.0-rc.2` automatically converts `xlink:href` to `href` when using `find()`.

```js
cheerio.load('')('<span><svg ><use xlink:href="#1" /></svg></span>').find('use').prop('xlink:href') //undefined
cheerio.load('')('<span><svg ><use xlink:href="#1" /></svg></span>').find('use').prop('href') //#1
```

Since, Safari doesn't support `href` without the "xlink" prefix in svg yet. Maybe, leave `xlink:href` as it was in `cheerio@^0.22.0` is better?

fb55:
Gives us a more capable serialiser.

This required the cloning logic to be replaced with lodash, as going through htmlparser2 will remove properties that parse5 relies on. That should also be considerably faster, so why not 😄

Fixes #1101


lib/utils.js
```diff
@@ -60,12 +58,21 @@ exports.domEach = function(cheerio, fn) {
...
+exports.cloneDom = function(dom) {
+  function filterOutParent(node) {
+    if (
+      "length" in dom
+        ? Array.prototype.some.call(dom, function(el) {
+          return node === el.parent;
+        })
+        : node === dom.parent
+    ) {
+      return null;
```
jugglinmike: This is tough to read syntactically, and the meaning is also a little hard to understand. What do you think about this:

```diff
+var parents = "length" in dom ?
+  Array.prototype.map.call(dom, function(el) { return el.parent; }) :
+  [dom.parent];
 function filterOutParent(node) {
-  if (
-    "length" in dom
-      ? Array.prototype.some.call(dom, function(el) {
-        return node === el.parent;
-      })
-      : node === dom.parent
-  ) {
+  if (parents.indexOf(node) > -1) {
     return null;
   }
 }
```

I have a feeling that `indexOf` will perform slightly better than `some`, but I'd recommend this even without that consideration.

lib/utils.js
```diff
@@ -60,12 +58,21 @@ exports.domEach = function(cheerio, fn) {
...
+    }
+  }
 
-  return parse(render(dom, options), options, false).children;
+  return cloneDeepWith(dom, filterOutParent);
```
jugglinmike: The documentation still describes the heuristic you're replacing. Could you update that,too?

jugglinmike:
This is cool!

fb55:
Thanks for the review, updated!

jugglinmike:
Merged to `v1.0.0` at commit e71d0ce6398c920a5c8d5b826f01f7a34da9d15a . Thanks, @fb55!

PR #1184, opened 2018-04-22
Formally test deprecated APIs

jugglinmike:
Okay, @fb55; here's the next step for gh-1122 :)

---

Methods named `load`, `html`, `xml`, and `text` are defined in many
locations:

Today, Cheerio defines multiple versions of methods named `load`,
`html`, `xml`, `text`, and `parseHTML`. These alternate versions may be
defined in up to three distinct parts of the API:

- exported by the Cheerio module
- as static methods of the "loaded" Cheerio factory function
- as instance methods of the "loaded" Cheerio factory function

Some of these are surperfluous, and because some unecessarily conflict
with idiomatic jQuery coding patterns, they have been designated for
future removal [1].

Add tests for the deprecated methods in order to avoid regressions prior
to their removal. Insert comments to delineate the methods which are
endorsed and which have been deprecated. For the latter group of
methods, include recommendation for the preferred alternatives.

[1] https://github.com/cheeriojs/cheerio/issues/1122

### DIFF
--- a/test/api/utils.js
+++ b/test/api/utils.js
@@ -21,33 +21,24 @@ describe('cheerio', function() {
       var $elem = cheerio('<span>foo</span>').html('');
       expect(cheerio.html($elem)).to.equal('<span></span>');
     });
-
-    it('() : of empty cheerio object should return null', function() {
-      expect(cheerio().html()).to.be(null);
-    });
-
-    it('(selector) : should return the outerHTML of the selected element', function() {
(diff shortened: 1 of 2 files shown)


jugglinmike:
@fb55 Do you have some time to look this over for me?

fb55:
Sorry for the delay — this looks dope!

jugglinmike:
Great, thanks! Mind doing the honors?

PR #1197, opened 2018-05-12
Implement for...of iterator via Symbol.iterator

### LINKED ISSUE #1191: Support for...of and entries iteration?
Hi!

I wanted to do some quick scraping and found `$(...).each(() => {...})` a bit uncomfortable now that we have `for ... of` in ES6. Basically I'd like to do the following:

```js
const $ = cheerio.load(myHtml);
for (const element of $('.someClass')) {
  // ...
}

for (const [index, element] of $('.someOtherClass').entries()) {
  // ...
}
```

I whipped up a quick patch that allows the above syntax:

```js

cheerio.prototype[Symbol.iterator] = function* () {
  for (let i = 0; i < this.length; i += 1) {
    yield this[i];
  }
};

cheerio.prototype.entries = function* () {
  for (let i = 0; i < this.length; i += 1) {
    yield [i, this[i]];
  }
};
```

LMK if this is out of scope -- if not I'll whip up a PR.

papandreou:
Similar to jQuery: https://github.com/jquery/jquery/blob/1ea092a54b00aa4d902f4e22ada3854d195d4a18/src/core.js#L371-L373

Fixes #1191


fb55:
This is a great way to implement this. Changed the base to the v1.0 branch, will merge this once the tests pass.

lib/cheerio.js
```diff
@@ -138,6 +139,11 @@ Cheerio.prototype.toArray = function() {
   return this.get();
 };
 
+// Support for (const element of $(...)) iteration:
+if (typeof Symbol !== 'undefined') {
+  Cheerio.prototype[Symbol.iterator] = Array.prototype[Symbol.iterator];
```
jugglinmike: This will fit better in `api/traversing.js`

test/api/traversing.js
```diff
@@ -680,6 +681,20 @@ describe('$(...)', function() {
...
+  if (typeof Symbol !== 'undefined') {
+    describe('[Symbol.iterator]', function() {
+
+      it('should yield each element', function() {
+        // The equivalent of: for (const element of $('li')) ...
+        var $li = $('li'),
+            iterator = $li[Symbol.iterator]();
+        expect(iterator.next().value.attribs['class']).to.equal('apple');
+        expect(iterator.next().value.attribs['class']).to.equal('orange');
+        expect(iterator.next().value.attribs['class']).to.equal('pear');
```
jugglinmike: Can you also assert that iteration terminates as expected?

papandreou: Absolutely, fixed in 1b80376066c8e0f272643857c5b5567ccb162aef

fb55:
Awesome, thanks a lot!

PR #1209, opened 2018-07-04
Fix running test on Windows

bzoz:
Moves lint and test script call from Makefile to package.json. Makes the test run on all platforms.


jdalton:
@jugglinmike This PR would allow unskipping `cheerio` in Node's [CitGM](https://github.com/nodejs/citgm) runs for Windows.

PR #1212, opened 2018-07-25
Updates htmlparser2 to latest version to improve compability issues with latest Node.js versions

SoftwareAndOutsourcing:
The latest published version of htmlparser2 seems having compatibility issues with the latest Node.js and/or some framework versions. This has been fixed in master (see https://github.com/fb55/htmlparser2/pull/233) so this PR updates the dependency to use the current git version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "css-select": "~1.2.0",
     "dom-serializer": "~0.1.0",
     "entities": "~1.1.1",
-    "htmlparser2": "^3.9.1",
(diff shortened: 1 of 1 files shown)


gajus:
This is fixed in 3.9.2.

SoftwareAndOutsourcing:
@gajus I tried 3.9.2 and I'm getting the same issues. 3.9.2 was released on Oct 2106 and the PR is from May 2018.

fb55:
This should have been fixed independently.

PR #1225, opened 2018-09-02
Add badge to display install size

styfle:
This adds a badge to the README.md file to display the npm install size


fb55:
It is not a goal of this project to have a small install size. Thanks for the patch, closing this as it does not seem very relevant to the project.

PR #1232, opened 2018-10-08
Refactor internals to designate deprecated API

jugglinmike:
Hey @fb55!

Taking your call to action for 1.0, here's the change set I had in mind for ["step 3" in gh-1122](https://github.com/cheeriojs/cheerio/issues/1122#issuecomment-383343689). It's a little larger than I anticipated back in April because while writing it, I found a couple more APIs that we should test (despite their need for removal): `cheerio.root` and `cheerio.contains`.

I'm hoping that with this in place, we should be able to generate documentation that fully captures all the deprecations we're making. It'll be my pleasure to follow up with the documentation generation process.


jugglinmike:
After thinking about this a little more, I'm concerned that future contributors may be confused by duplicated tests for methods with the same name. Given the amount of confusion these duplicated method names have caused us already, it seems likely that someone is going to mistakenly change the wrong tests. That's why I've pushed up another commit to move all the tests for deprecated APIs into a dedicated file. This should help avoid confusion, and it will also simplify the process of finally removing the methods.

jugglinmike:
Would you merging on my behalf? Sorry to be a stickler, I just prefer to have the review process reflected in the git history.

fb55:
Sure thing!

jugglinmike:
Wowza! Thanks :D

matthewmueller:
looks great!!

PR #1237, opened 2018-10-14
Polish the procedurally-generated documentation website

jugglinmike:
This shuffles some of the source around so that we can generate descriptive docs. It includes a kind of ugly workaround (`exports = exports;`), but we've gone without human-readbale docs for too long, so I'm willing to be pragmatic.

As noted in the commit message, one technical issue can't be worked around with code formatting, so [I patched JSDoc](https://github.com/cheeriojs/jsdoc/commit/1aa9563256613863c2fd6cc489f3887f5c6df9b4). Hopefully, [they'll accept the patch upstream](https://github.com/jsdoc3/jsdoc/pull/1575), but for now, it's available on [a fork of the project owned by the "cheeriojs" GitHub organization](https://github.com/cheeriojs/jsdoc) (just in case we need to make more changes in the future).

I've also included a modified version of JSDoc's default template to suit the needs of this project. I did this by first committing the default template without any modifications. That way, our customizations are more evident in the project's git history. Here are a few screen captures to help give folks an idea of what to expect.

The home page is the project's readme file:

2018-10-08-docs-home

The word "Categories" in the menu is a customization. Technically, we're using JSDoc's "mixins" to model the relationship of Cheerio's internal modules. We shouldn't document them in that way, though, as the term "mixin" has some precedence in the context of JavaScript modules, but it's not actually relevant here. We should avoid suggesting usage patterns that we don't support, so I chose "Category" as a more neutral term--one that's also used in [the jQuery API documentation](https://api.jquery.com).

That's also used when describing the Cheerio object:

2018-10-08-docs-categories

The last customization isn't technically necessary. By default, any methods marked with the `@deprecated` sigil were simply rendered with the word "Deprecated" inserted at the end of their description. I've further de-emphasized their presentation with a `<details>` element:

2018-10-08-docs-deprecated


fb55:
@jugglinmike Fixed the tests, feel free to merge :)

fb55:
Decided to merge this to move things along.

PR #1246, opened 2018-11-01
Updated parse5 to 4.0.0 (fixes #1245).

### LINKED ISSUE #1245: Dependencies on cheerio 1.0.0-rc2 in an ES3 TypeScript project will cause build failures
cheerio 1.0.0-rc2 has a dependency on parse5 3.0.1, which itself has a dependency on types/node.  The result of this is that installing cheerio installs types/node, which means that the TypeScript compiler uses those types when compiling, even if the target is ES3, causing compile errors ([similar to this issue](https://github.com/DefinitelyTyped/DefinitelyTyped/issues/24595)).

Tests failed on updating parse5 to the latest version of 5.1.0, but I was able to update it to 4.0.0 without a problem (PR on the way).

jameswilddev:
Tests run successfully.  This is specifically to fix projects including cheerio failing to compile as ES3 TypeScript (see #1245), which [seems to be what 3.0.1->4.0.0 addresses](https://github.com/inikulin/parse5/blob/master/docs/version-history.md#400).

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "entities": "~1.1.1",
     "htmlparser2": "^3.9.1",
     "lodash": "^4.17.5",
-    "parse5": "^3.0.1"
+    "parse5": "^4.0.0"
   },
   "devDependencies": {
     "benchmark": "^2.1.0",


fb55:
Good find, thanks a lot!

chigix:
When will this commit be released? I need this commit

PR #1248, opened 2018-11-03
Remove references to deprecated APIs from docs

jugglinmike:
@fb55 Even with this patch applied, the readme will continue to encourage the
use of `$.xml`. Do you know how to rewrite that in terms of `cheerio.xml`?

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -233,21 +233,26 @@ $('[xml\\:id="main"');
 ```
 
 ### Rendering
-When you're ready to render the document, you can use the `html` utility function:
+When you're ready to render the document, you can call the `html` method on the "root" selection:
 
 ```js
-$.html()
-//=>  <ul id="fruits">
-//      <li class="apple">Apple</li>
-//      <li class="orange">Orange</li>
-//      <li class="pear">Pear</li>
-//    </ul>
+$.root().html()
+//=>  <html>
+//      <head></head>
+//      <body>
+//        <ul id="fruits">
+//          <li class="apple">Apple</li>
+//          <li class="orange">Orange</li>
+//          <li class="pear">Pear</li>
+//        </ul>
+//      </body>
+//    </html>
 ```
 
-If you want to return the outerHTML you can use `$.html(selector)`:
+If you want to render the [`outerHTML`](https://developer.mozilla.org/en-US/docs/Web/API/Element/outerHTML) of a selection, you can use the `html` utility functon:
 
 ```js
-$.html('.pear')
+cheerio.html($('.pear'))
 //=> <li class="pear">Pear</li>
 ```
 
@@ -268,15 +273,7 @@ You may also render the text content of a Cheerio object using the `text` static
 
 ```js
 const $ = cheerio.load('This is <em>content</em>.')
-$.text()
-//=> This is content.
-```
-
-The method may be called on the Cheerio module itself--be sure to pass a collection of nodes!
-
-```js
-const $ = cheerio.load('<div>This is <em>content</em>.</div>')
-cheerio.text($('div'))
+cheerio.text($('body'))
 //=> This is content.
 ```
 


fb55:
Finally got around to having a look at this. Shouldn't `xml` be equivalent to `html`?

jugglinmike:
@fb55 Here's the text in question:

> By default, `html` will leave some tags open. Sometimes you may instead want
> to render a valid XML document. For example, you might parse the following
> XML snippet:
>
> ```xml
> const $ = cheerio.load('<media:thumbnail url="http://www.foo.com/keyframe.jpg" width="75" height="50" time="12:05:01.123"/>');
> ```
>
> ... and later want to render to XML. To do this, you can use the 'xml' utility function:
>
> ```js
> $.xml()
> //=>  <media:thumbnail url="http://www.foo.com/keyframe.jpg" width="75" height="50" time="12:05:01.123"/>
> ```

This patch updated the recommendation about `$.html` (using `cheerio.html`
instead). If we make a similar recommendation about `$.xml`, users won't get
the expected result.

Here's the output in v0.22.0:

    $.xml()
    //=> <media:thumbnail url="http://www.foo.com/keyframe.jpg" width="75" height="50" time="12:05:01.123"/>
    cheerio.xml($.root())
    //=> <media:thumbnail url="http://www.foo.com/keyframe.jpg" width="75" height="50" time="12:05:01.123"/>

In v1.0.0, the source text has already been parsed as HTML, and that's
destructive. Passing a selection created in this way will not produce XML:

    $.xml()
    //=> <html><head></head><body><media:thumbnail url="http://www.foo.com/keyframe.jpg" width="75" height="50" time="12:05:01.123"></media:thumbnail></body></html>
    cheerio.xml($.root())
    <root><html><head></head><body><media:thumbnail url="http://www.foo.com/keyframe.jpg" width="75" height="50" time="12:05:01.123"></media:thumbnail></body></html></root>

`cheerio.html` accepts a Cheerio object and returns a string representation of
the markup. `cheerio.xml` can't necessary do the same because the parsing has
already taken place.

jugglinmike:
@fb55 You may have missed my previous comment, but I'm still not sure how to address it. Could you take a look?

fb55:
Hi Mike, sorry, didn't look at this for a while. I am unable to replicate your result.  We do get the wrapper, but we never render the root tag.

```js
$.xml()
//=> <html><head></head><body><media:thumbnail url="http://www.foo.com/keyframe.jpg" width="75" height="50" time="12:05:01.123"></media:thumbnail></body></html>
cheerio.xml($.root())
//=> <html><head></head><body><media:thumbnail url="http://www.foo.com/keyframe.jpg" width="75" height="50" time="12:05:01.123"></media:thumbnail></body></html>
```

We should perhaps mention that parsing should be done with `xmlMode: true` to avoid getting a wrapper. 

```js
$ = cheerio.load('<media:thumbnail url="http://www.foo.com/keyframe.jpg" width="75" height="50" time="12:05:01.123"/>', {xmlMode:true});
$.xml()
// => '<media:thumbnail url="http://www.foo.com/keyframe.jpg" width="75" height="50" time="12:05:01.123"/>'
```

jugglinmike:
I see the same output that you have reported if I use the latest release candidate (rc3), but I still see the output I shared if I use the latest version of the `v1.0.0` branch.

This is a bit confusing, but the `v1.0.0` branch is actually *ahead* of the release candidate because [we published rc3 as a hotfix for rc2](https://github.com/cheeriojs/cheerio/pull/1291).

    $ git checkout master
    Already on 'master'
    $ git rev-parse HEAD
    7a9ff22f68bd6dbdf04f8e3a52e28a6b369b5fa7
    $ node -e "c = require('./'); $ = c.load('<media:thumbnail/>'); console.log(c.xml($.root()));"
    <media:thumbnail/>
    $ git checkout v1.0.0-rc.3 
    Switched to branch 'v1.0.0-rc.3'
    $ git rev-parse HEAD
    cce5c1878fc3fa9c194c88fefcbc216d654fabaa
    $ node -e "c = require('./'); $ = c.load('<media:thumbnail/>'); console.log(c.xml($.root()));"
    <html><head></head><body><media:thumbnail></media:thumbnail></body></html>
    $ git checkout v1.0.0
    Switched to branch 'v1.0.0'
    $ git rev-parse HEAD
    bfd84be09ecf76faed9fbddd39fd8559461f343f
    $ node -e "c = require('./'); $ = c.load('<media:thumbnail/>'); console.log(c.xml($.root()));"
    <root><html><head></head><body><media:thumbnail></media:thumbnail></body></html></root>

In addition to the wrapper, I'm concerned about the insertion of the closing tag. As I understand it, this is because we've already parsed the input as HTML, and parse5 has automatically inserted the closing tag as per the HTML parsing spec. Later requests to render the tree as XML contain the tag.

This seems like a potential regression. When we used htmlparser2 consistently, we parsed in a way that was lax and tolerated XML. The internal data structure could be rendered as XML without any loss in fidelity. Now, all input will be parsed as HTML to-spec, which (as I understand it) involves inserting the closing tag.

Having written all this out, a partial solution occurs to me. We agreed to continue to support XML (via `htmlparser2`) with an opt-in passed to `cheerio.load`. We could just update the example I've referenced accordingly, and agree that the old "parse as both HTML and XML" behavior is something we're removing as a breaking change.

Even then, though, it seems as though `htmlparser2` adds the closing tag:

    $ node -e "c = require('./'); $ = c.load('<media:thumbnail/>', { xml: true }); console.log(c.xml($.root()));"
    <root><media:thumbnail></media:thumbnail></root>

Does that surprise you, too?

jugglinmike:
I think I've solved the inconsistency with the wrapper--see gh-1307.

I still don't understand why/how htmlparser2 is inserting the closing tag, though.

PR #1249, opened 2018-11-06
add option keep non ascii

GHolk:
a few little patch include **dom-serializer** , **entities** ,
allow output keep non ascii character raw.

```
const cheerio = require('cheerio')

cheerio.load('<h1>你好 :)</h1>').html()
// <h1>&#x4F60;&#x597D; :)</h1>

cheerio.load('<h1>你好 :)</h1>', {keepNonAscii: true}).html()
// <h1>你好 :)</h1>
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -269,7 +269,8 @@ of the default parsing options:
 ```js
 const $ = cheerio.load('<ul id="fruits">...</ul>', {
     normalizeWhitespace: true,
-    xmlMode: true
+    xmlMode: true,
+    keepNonAscii: true
 });
 ```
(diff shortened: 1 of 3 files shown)


dzcpy:
This problem obviously needs to be fixed, but in package.json, referencing a module from git repo is not ideal.

GHolk:
i reference module in github because it is a temparory fix. if dom-serializer accept my pull request, i can change dependency to the dom-serializer on npm.

fb55:
The current 1.0.0 branch behaves exactly this way.

PR #1254, opened 2018-12-03
Fix #1253 (stale data on .prev() after replaceWith())

### LINKED ISSUE #1253: Stale data in prev sibling after replaceWith()
I called `.replaceWith()` on a node with one sibling before and one sibling after. Then:

- Getting that node from the prev sibling using `.next()` returns the new node (correct behavior).
- Getting that node from the next sibling using `.prev()` returns the old (replaced) node.

Here's the code: https://runkit.com/embed/cspja1n2yh30

In jquery, it works as expected: https://codepen.io/gei0r/pen/yQwpeG

    let $ = cheerio.load(`
    <div id="root">
      <div id="0">Zero</div>
      <div id="1">Stale value, will be replaced</div>
      <div id="2">Two</div>
    </div>`);
    
    let zero = $('#0');
    let two = $('#2');
    
    $('#1').replaceWith('<div id="3">New value: This is ok</div>');
    
    console.log(zero.next().html());
    console.log(two.prev().html());

Output:

    "New value: This is ok"
    "Stale value, will be replaced"

Is there at least a way to "re-scan" the DOM so that the stale data is refreshed?

Gei0r:
Hi,
I fixed the error I described in #1253. I also modified two test cases to demonstrate the fix.

The problem was that in `uniqueSplice()`, the variable  `next` was calculated wrongly:
- Before: `next` pointed at `spliceIdx`, which actually containted the original node to be replaced. When it came to updating references in the linked list (line 104), the reference in the replaced element was changed. But the reference in the following element was not updated so it still pointed to the replaced element.
- After: Now the `prev` reference in the element *following* the newly inserted element(s) is updated.


fb55 (approved):
Nice catch, thanks for this!

PR #1293, opened 2019-03-25
Fix buffer deprecation warning

BridgeAR:
This makes sure the test uses `Buffer.from` instead of `new Buffer`
to prevent the deprecation warning. It will work with all Node.js
versions due to using the `safe-buffer` module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,10 +36,11 @@
     "coveralls": "^2.11.9",
     "expect.js": "~0.3.1",
     "istanbul": "^0.4.3",
-    "jsdom": "^9.2.1",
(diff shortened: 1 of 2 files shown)


BridgeAR:
I just saw that @lpinca opened a PR to fix this a long time ago: https://github.com/cheeriojs/cheerio/pull/1234.

@jugglinmike @fb55 it would be really great if you could merge that PR! We track this deprecation warning in modules and try to resolve it everywhere.

fb55:
Merged #1234! Thanks for flagging this!

Issue #1199, opened 2018-05-15
No lockfile?


bencooper222:
This package doesn't contain a `yarn.lock` or `package-lock.json`. Is there any design reason for this? Lockfiles should generally be pushed because they give more [consistent builds](https://yarnpkg.com/blog/2016/11/24/lockfiles-for-all/).

fb55:
The last release was before npm introduced `package-log.json` files. Future releases will probably feature them, just hasn't been a priority so far.

Issue #1175, opened 2018-04-01
DoS/RCE vulnerability in lodash@4.15.0


voxxit:
See: https://www.sourceclear.com/vulnerability-database/security/prototype-pollution/javascript/sid-5818

**Fix:** Upgrade to `lodash@^4.17.5`.

fb55:
PRs welcome! :)

jodylecompte:
@fb55 I've opened up a PR to correct this. I didn't see any guidance on branches in the contributing doc but it looked like the last few PRs were branched off of v1.0.0 so I did this as well. Let me know if this was not correct and I'll fix it.

fb55:
Thanks @jodylecompte!

PR #1300, opened 2019-04-17
update dependency to resolve security issue

compilenix:
https://github.com/lodash/lodash/issues/3359


fb55:
Thanks!

compilenix:
One questtion @fb55 :
Would it be possible to add a package-lock.json any time soon?

Reference: https://github.com/cheeriojs/cheerio/issues/1199#issuecomment-389305568

fb55:
Two reasons why it was removed: 1) They are hard to review and can point to random files on the internet, which I would consider a security risk and 2) similarly, dealing with merge conflicts is a pain.

jugglinmike:
I'm a little late to the party, but this patch was superfluous.

The Lodash vulnerability reported here was previously reported in gh-1175 and
subsequently fixed via gh-1179. The reason the fix was not available in
Cheerio's "rc3" release is that `v1.0.0-rc.3` was created as a hotfix for
`v1.0.0-rc.2`--it does not have any of the changes which occured in `master`
since the second release candidate.

So while the next release of Cheerio will not include this commit, neither will
it include the vulnerable version of Lodash. In order to avoid further
confusion, I will delete the branch named `v1.0.0-rc3`. We will continue to
maintain the corresponding *tag* named `1.0.0-rc3` for the purposes of
archiving,

PR #1307, opened 2019-05-13
Correct rendering of root node

jugglinmike:
This change normalizes rendering between trees generated by parse5 and
htmlparser2. Consistent rendering of the "root" node is especially
important given that `cheerio.render($.root())` will be the recommended
pattern for rendering entire documents in Cheerio version 1.0.


jugglinmike:
Sure thing

PR #1334, opened 2019-07-23
removeAttr just only removes one attr 

### LINKED ISSUE #1316: removeAttr just only remove one attr
removeAttr just only remove one attr
can't like jq use space remove more attr

16546513:
fix #1316


fb55:
Closing in favor of #1561

PR #1411, opened 2020-05-30
Chinese readme link added on main readme

frankda:
As Felix required, added a markdown link with English description to the new URL of the Chinese readme article.


fb55:
Thanks a lot @frankda!

PR #1486, opened 2020-09-18
[ ADD ] Support for react-native

grovertb:


package.json
```diff
@@ -28,6 +28,7 @@
     "css-select": "~2.1.0",
     "dom-serializer": "~1.1.0",
     "entities": "~2.0.2",
+    "events": "^3.2.0",
```
fb55, line 31: `events` isn't used anywhere in this module. Let's add this to the module actually pulling in the dependency, or have users add it to their own `package.json`s.

I know that `htmlparser2` currently depends on it, but that will be fixed before the next cheerio release.

fb55 (commented):
Checking if `Buffer` exists is a good idea. Happy to merge if we can remove the `events` change.

package.json
grovertb: I just removed the package "events"

fb55:
Thanks a lot @grovertb!

Issue #1471, opened 2020-08-31
Consider owning @types/cheerio type definitions


matthewmueller:
### Problem

In the early days, jQuery familiarity was one of the main attractions to Cheerio. Nowadays it may be more of a hindrence.

For example, it's confusing that Cheerio's `$(selector).each(fn)` has a different argument order than ES5's `forEach`.

```js
const fruits = [];

$('li').each(function(i, elem) {
  fruits[i] = $(this).text();
});

fruits.join(', ');
//=> Apple, Orange, Pear
```

### Potential Solution

One way to reduce the friction is to have autocomplete. We currently support some autocomplete via JSDoc but the type definitions would take it a step further.

Some wonderful folks have been maintaining [@types/cheerio](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/3e0007a3a88eca10f61fcb55d78dd66e323f24c6/types/cheerio/index.d.ts). Installing this extra package is quite obvious for Typescript users but less obvious for Javascript developers. 

We could make this simpler by hosting those types within the Cheerio repository. This would be one less thing to install and the type definitions would get picked up for Javascript users.

### Downsides

- The maintenance burden is now on the Cheerio maintainers. 
- Any other downsides?

bbqsrc:
Writing a scraper, as I am wont to do at least once per year, and straight up came here to ask why it wasn't hosted in the repo yet.

Timing was good, it seems. :smile:

fb55:
@AzSiAz Afaict you seem to be the person most on top of the current type definitions (thank you or that!). Any concerns about us pulling the type definitions into the cheerio repo? Anyone else you think we should loop in?

AzSiAz:
@fb55 Thanks for asking, glad you're a considering owning the type definition :smile:

Maybe the other typing maintainer so they know if they need to make modification but I think most of them don't participate anymore so do as you see fit on that front

The only concern coming to mind is to be careful about the node definition included in the current types, you may have to remove/"polyfill" it instead since many people seem to have a problem with it, and it's not really a concern but there is also a PR open [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/46006): you might want to check with the author to get change concerning cheerio integrated 

Also, don't forget to delete types on definitelyTyped and update types depending on cheerio, might be a pain thought^^

fb55:
Thanks a lot @AzSiAz, this is very helpful.

[...]

PR #1491, opened 2020-09-25
Pull in types from DefinitelyTyped

paulmelnikow:
This is an attempt to pull in the Cheerio types from DefintelyTyped so that they're available to all JS users, and so TS + JS users are assured that the type definitions match the version of the library they are using without having to determine the correct version of the @types/cheerio package.

It's the first time I've done this, so feedback from knowledgeable people would be very welcome.

I have two specific questions:

1. I've added `@types/node` to dependencies because the types use `Buffer`. Does the library target the browser and React Native? If so will including `@types/node` as a dependency cause problems? Is it preferable to use [a reference](https://www.typescriptlang.org/docs/handbook/triple-slash-directives.html)?
2. In DT, the type defs close with `declare module 'cheerio'`. https://github.com/DefinitelyTyped/DefinitelyTyped/blob/fe3e8016d42f86fff607a7cd424ec3cabeea4b45/types/cheerio/index.d.ts#L304-L307 That wasn't working, so I've changed it here to `export = cheerioModule` which seems to work. Is there a better way to do that?

Before this is merged and shipped we should definitely test that the types can be used successfully by a dependent package. After this is shipped, the @types/cheerio package should be deprecated, and the DT types which depend on Cheerio updated accordingly. I'm not totally sure how to do that, but if someone can point me in the right direction I'm happy to do it.

/cc @matthewmueller @ljharb

Ref #1471

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,14 +17,17 @@
     "url": "git://github.com/cheeriojs/cheerio.git"
   },
   "main": "./index.js",
+  "types": "types/index.d.ts",
   "files": [
     "index.js",
+    "types/index.d.ts",
     "lib"
   ],
   "engines": {
     "node": ">= 0.6"
   },
   "dependencies": {
+    "@types/node": "^14.11.2",
     "css-select": "~2.1.0",
     "dom-serializer": "~1.1.0",
     "entities": "~2.0.2",
@@ -48,12 +51,14 @@
     "mocha": "^8.1.1",
     "nyc": "^15.0.1",
     "prettier": "^2.1.1",
+    "tsd": "^0.13.1",
     "xyz": "~4.0.0"
   },
   "scripts": {
     "test": "npm run test:lint && npm run test:mocha",
     "test:mocha": "mocha --recursive --reporter dot --parallel",
     "test:lint": "eslint --ignore-pattern docs --ignore-path .gitignore .",
+    "test:types": "tsd",
     "pre-commit": "lint-staged"
   },
   "prettier": {
--- a/Makefile
+++ b/Makefile
@@ -5,6 +5,9 @@ UPSTREAM = git@github.com:cheeriojs/cheerio.git
 lint:
 	@npm run test:lint
 
+types:
+	@npm run test:types
+
 test:
 	@npm run test
 
@@ -23,7 +26,7 @@ report-cov: test-cov
 	@./node_modules/.bin/nyc report --reporter=text-lcov | ./node_modules/.bin/coveralls || [ "$(OPTIONAL)" = "true" ]
 
 travis-test: OPTIONAL = true
-travis-test: lint report-cov
+travis-test: lint types report-cov
 	@true
 
 bench:
--- a/package-lock.json
+++ b/package-lock.json
@@ -322,10 +322,68 @@
       "integrity": "sha512-tsAQNx32a8CoFhjhijUIhI4kccIAgmGhy8LZMZgGfmXcpMbPRUqn5LWmgRttILi6yeGmBJd2xsPkFMs0PzgPCw==",
       "dev": true
     },
+    "@nodelib/fs.scandir": {
+      "version": "2.1.3",
+      "resolved": "https://registry.npmjs.org/@nodelib/fs.scandir/-/fs.scandir-2.1.3.tgz",
+      "integrity": "sha512-eGmwYQn3gxo4r7jdQnkrrN6bY478C3P+a/y72IJukF8LjB6ZHeB3c+Ehacj3sYeSmUXGlnA67/PmbM9CVwL7Dw==",
+      "dev": true,
+      "requires": {
+        "@nodelib/fs.stat": "2.0.3",
+        "run-parallel": "^1.1.9"
+      }
+    },
+    "@nodelib/fs.stat": {
+      "version": "2.0.3",
+      "resolved": "https://registry.npmjs.org/@nodelib/fs.stat/-/fs.stat-2.0.3.tgz",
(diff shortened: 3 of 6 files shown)


package.json
```diff
@@ -17,14 +17,17 @@
...
   "files": [
     "index.js",
+    "types/index.d.ts",
     "lib"
   ],
   "engines": {
     "node": ">= 0.6"
   },
   "dependencies": {
+    "@types/node": "^14.11.2",
```
ljharb, line 30: adding TS things to deps is a *very bad idea* since the vast majority of consumers are not using TS.

ljharb (changes requested):

ljharb:
Specifically, shipping types with a package is also a very bad idea - types should only EVER be shipped via DT.

fb55:
> shipping types with a package is also a very bad idea - types should only EVER be shipped via DT

Modules written in TS have always shipped with types, I don't think this is out of the norm. Any particular reason why we shouldn't ship types with cheerio?

Regarding the @types/node dependency: Ideally we'd work around pulling this in for all users. One possible solution would be to use `Uint8Array` as the type, as `Buffer` extends this nowadays. This is the approach eg. [uWebSockets](https://github.com/uNetworking/uWebSockets.js) follows.

ljharb:
Cheerio isn’t authored in typescript though.

One huge problem with it is that you conflate two APIs into the same semver. Needing to make a major bump to cheerio in order to make a breaking change to the types has very bad downstream effects. There’s a reason yargs 15 started shipping types and 16 removed them.

paulmelnikow:
An advantage of shipping the types with the library is it's easier for downstream users to match the types to the version of the library they are actually using. That's been a plus for nock. Breaking changes to types in DT aren't even tracked with semver (i.e. they ship in what would be patch releases) so I'm not convinced DT provides a perfect solution to the question of whether to version the JS API or the types. I'd be interested in the discussions that led to the decision for yargs.

> Needing to make a major bump to cheerio in order to make a breaking change to the types has very bad downstream effects.

What are the bad effects?

ljharb:
A semver major bump discourages updates, because of the risk of migration effort.

TS users have chosen “not JavaScript” and as such, they have signed up for the extra effort that entails. It is unacceptable to burden JavaScript users because some people choose to use an alternative language.

paulmelnikow:
> A semver major bump discourages updates, because of the risk of migration effort.

A changelog helps with this by clarifying what is affected by the semver major.

> TS users have chosen “not JavaScript” and as such, they have signed up for the extra effort that entails. It is unacceptable to burden JavaScript users because some people choose to use an alternative language.

I'm not sure it's as simple as this. The maintainer's motivation for this PR in #1471 was to improve the VS Code autocompletion for JavaScript users. Maintainers' time should be considered too. As the API changes, maintaining the types within the project is easier than doing it in DT because the changes can be made and tested in the same PR. Shipping the types here is especially important for the contributor workflow for TS users. Developing on or maintaining a fork is much harder when the types are external.

I get that there are some negatives to shipping the types within packages, including meaningful negatives for some users who might not ever care about TS, however completely insulating these users from the fact that TS exists seems short-sighted, given the easier maintenance and significant benefits for TS users and many JS users who unknowingly use the TS-driven autocompletion in VS Code.

ljharb:
Wait, if the project isn't authored in TS, how are the types tested in this project?

paulmelnikow:
As in DT, this PR checks in tests. In PRs that change the interface, the type changes are tested too.

When a TS project uses Cheerio as a dependency, and modifies the library and its interface, the type changes are necessary for testing the library changes within the context of the project.

fb55:
Thanks a lot @paulmelnikow and @ljharb for finding an agreement here and getting to a better solution in the process!

Quite a few of the underlying packages for this module already have type definitions. Would it make sense to eg. replace the `Element` class with the exported class from `domhandler`?

ljharb:
@fb55 to be clear, I still don't think types should be shipped in a non-TS package, but it's also not up to me.

paulmelnikow:
> @fb55 to be clear, I still don't think types should be shipped in a non-TS package,

Yup, hope I didn't imply we'd somehow resolved our differences here 😁

paulmelnikow:
> Quite a few of the underlying packages for this module already have type definitions. Would it make sense to eg. replace the `Element` class with the exported class from `domhandler`?

Hmm, yes, it seems possible that would be better and seems a little odd that wasn't done before. I see cheerio requires htmlparser2 `^3` which transitively requires domhandler `^3`. Since all those types are shipped with those libraries, it's straightforward to depend on them.

Does cheerio have a way of shipping beta releases? The types I've pulled in here are identical to the last release in DT, though if we're going to make breaking changes to them, it probably would be good to validate them through beta releases.

fb55:
> Does cheerio have a way of shipping beta releases?

The last couple of releases were RCs, which should have been beta versions realistically. I'm okay with making breaking changes to the types before the final `1.0.0` release.

> I still don't think types should be shipped in a non-TS package

That does make sense. For context: I'd like to transition cheerio to TS eventually, and having the type definitions ready will be helpful.

paulmelnikow:
> The last couple of releases were RCs, which should have been beta versions realistically. I'm okay with making breaking changes to the types before the final `1.0.0` release.

Ahhh gotcha! I didn't realize cheerio was already in the midst of 1.0 prereleases.

Would you like me to update this to use the upstream types?

fb55:
I didn't really consider the fact that we have two different parsers in my earlier comments. Let's go with what we have right now, and adjust as we go. Thanks a lot @paulmelnikow for following through here!

PR #1499, opened 2020-10-01
Fixing lodash vulnerability

eklem:
What about this being v1.0.0-rc.4 ?

Test result (same as before lodash version bump):
```
  650 passing (1s)
  1 pending
```

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "cheerio",
-  "version": "1.0.0-rc.3",
+  "version": "1.0.0-rc.4",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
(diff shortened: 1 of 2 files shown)


fb55:
Thanks @eklem!

fb55:
I would want to include updates to my modules in the RC (that would be htmlparser2, as well as the new cheerio-select module). I'm wrapping up the last changes, the next RC should land soon after.

eklem:
Any progress on the `htmlparser2`and `cheerio-select`, @fb55 ?

fb55:
https://github.com/cheeriojs/cheerio-select/issues/2 is currently blocking. If this doesn't get resolved in the next two weeks, I will release the package under a different name.

PR #1500, opened 2020-10-01
Avoid lodash where possible

TrySound:
Lodash is quite big package https://packagephobia.com/result?p=lodash
and often used where modern javascript feature do a good job.

In this diff I replaced lodash with native functions or for/of loop.

The only left case is lodash/cloneDeepWith which does not have clear
alternative. I will try to replace it with something in another PR and
remove lodash dependency.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,14 +6,6 @@
   "extends": ["eslint:recommended", "prettier"],
   "globals": {},
   "rules": {
-    "indent": [
-      2,
-      2,
-      {
-        "SwitchCase": 1,
-        "VariableDeclarator": 2
-      }
-    ],
     "no-eq-null": 0,
     "no-proto": 2,
 
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -6,6 +6,14 @@
   "extends": ["eslint:recommended", "prettier"],
   "globals": {},
   "rules": {
+    "indent": [
+      2,
+      2,
+      {
+        "SwitchCase": 1,
+        "VariableDeclarator": 2
+      }
+    ],
     "no-eq-null": 0,
     "no-proto": 2,
 
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -8,25 +8,25 @@ exports = exports; // eslint-disable-line no-self-assign
 // The preceeding statement is necessary for proper documentation generation.
 
 var text = require('../static').text,
-  utils = require('../utils'),
-  isTag = utils.isTag,
-  domEach = utils.domEach,
-  hasOwn = Object.prototype.hasOwnProperty,
-  camelCase = utils.camelCase,
-  cssCase = utils.cssCase,
-  rspace = /\s+/,
-  dataAttrPrefix = 'data-',
-  // Lookup table for coercing string data-* attributes to their corresponding
-  // JavaScript primitives
-  primitives = {
-    null: null,
-    true: true,
-    false: false,
-  },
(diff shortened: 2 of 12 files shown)


TrySound:
cc @fb55

lib/api/css.js
```diff
@@ -85,7 +82,13 @@ function getCss(el, prop) {
   if (typeof prop === 'string') {
     return styles[prop];
   } else if (Array.isArray(prop)) {
-    return _.pick(styles, prop);
+    var newStyles = {};
+    for (const item of prop) {
```
fb55, line 86: How about

```suggestion
    prop.reduce(...)
```

lib/api/attributes.js
```diff
@@ -424,7 +419,7 @@ exports.removeAttr = function (name) {
  * @see {@link http://api.jquery.com/hasClass/}
  */
 exports.hasClass = function (className) {
-  return _.some(this, function (elem) {
+  return Array.from(this).some(function (elem) {
```
fb55, line 422: Can we use `toArray` instead of `Array.from`? (Same in other spots.)

```suggestion
  return this.toArray().some(function (elem) {
```

lib/api/traversing.js
```diff
@@ -36,13 +29,10 @@ var select = require('css-select'),
...
 exports.find = function (selectorOrHaystack) {
-  var elems = _.reduce(
-    this,
-    function (memo, elem) {
-      return memo.concat(_.filter(elem.children, isTag));
-    },
-    []
-  );
+  var elems = [];
+  for (const elem of Array.from(this)) {
```
fb55, line 33: `this` is already an iterable.

```suggestion
  for (const elem of this) {
```

fb55 (commented):
Great PR, thanks for looking into this! Left some comments, there are some patterns that can be simplified. Also, it would be great if we could avoid using iterators (both with `for .. of` loops and with `Array.from`, as some users are still using older JS engines. Note that this isn't forever & we will upgrade eventually.

TrySound:
I relied on travis.yaml and thought only node lts are supported. Also what about arrow functions?

TrySound:
@fb55 I converted all changes to es5

TrySound:
Though I left Object.assign as there is no alternative in es5

fb55:
> The only left case is lodash/cloneDeepWith which does not have clear alternative. I will try to replace it with something in another PR and remove lodash dependency.

I've [just added](https://github.com/fb55/domhandler/commit/5babe0dd8e6bb45c422d2daea5419abefe400dc9) a `cloneNode` method to `domhandler`, which should remove the need for this method altogether.

.eslintrc.json
```diff
@@ -6,14 +6,6 @@
   "extends": ["eslint:recommended", "prettier"],
   "globals": {},
   "rules": {
-    "indent": [
```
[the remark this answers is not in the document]
TrySound: You have prettier enabled. And its formatting is different than eslint one. I needed to pass eslint to test properly.
https://github.com/cheeriojs/cheerio/blob/v1.0.0/package.json#L59-L62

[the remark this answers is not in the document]
TrySound: I see. Pre-commit package is missing.

TrySound:
@fb55 Done

fb55:
Awesome work, thank you so much!

TrySound:
Hi @fb55. I don't see cloneNode available in cheerio nodes. Am I missing something?

PR #1559, opened 2020-12-11
refactor: Use the parser's DOM without modifications

fb55:
BREAKING: This removes the `root` property from top-level nodes. Instead, the `root` reference is kept in the (previously `null`) `parent` property. Cheerio now uses the root object generated by either of the parsers.

Also bumps `css-select`, as the new version includes a fix for `parent` references that aren't elements.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -182,7 +182,6 @@ The options in the `xml` object are taken directly from [htmlparser2](https://gi
 
 ```js
 {
-    withDomLvl1: true,
     normalizeWhitespace: false,
     xmlMode: true,
     decodeEntities: true
--- a/test/api/utils.js
+++ b/test/api/utils.js
@@ -33,7 +33,7 @@ describe('cheerio', function () {
       var $ = cheerio.load('<div>a div</div><span>a span</span>');
       var $collection = $('div').add($.root()).add('span');
       var expected =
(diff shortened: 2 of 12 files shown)


Readme.md
```diff
@@ -182,7 +182,6 @@ The options in the `xml` object are taken directly from [htmlparser2](https://gi
 
 ```js
 {
-    withDomLvl1: true,
```
fb55, line 185: This option no longer exists in `htmlparser2`.

test/api/utils.js
```diff
@@ -33,7 +33,7 @@ describe('cheerio', function () {
       var $ = cheerio.load('<div>a div</div><span>a span</span>');
       var $collection = $('div').add($.root()).add('span');
       var expected =
-        '<span>a span</span><html><head></head><body><div>a div</div><span>a span</span></body></html><div>a div</div>';
```
fb55, line 36: These elements used to be disjoint, because `uniqueSort` did not resolve the `root` reference. Now, `uniqueSort` will actually put the document first & the span last (as it should).

5saviahv:
If root element is removed does parse5 htmlparser2 tree-adapter also needs to change ?? So it wont generate root node ?

fb55:
Both parsers produce root nodes. This change prevents cheerio from taking away the existing root nodes and replacing them with a custom one.

Updating the description above as I can see where the confusion comes from.

PR #1564, opened 2020-12-15
refactor: Use `domhandler` nodes directly

### LINKED ISSUE #1166: Improve performance of `.text()`
We recently improved the correctness of the `text()` method in a way that degrades performance. Restoring the performance of that method will require a new feature in the domhandler project.

See gh-588 for more detail.

### LINKED ISSUE #1506: Replacing cloneDeepWidth with alternative
[Here](https://github.com/cheeriojs/cheerio/pull/1500#issuecomment-702999040) was mentioned new cloneNode api in domhandler. Though it covers only htmlparser2 integration. Is there any similar api in parse5?

cc @fb55

fb55:
Avoids unnecessary parser calls. Fixes #1166.

Allows us to drop lodash. Fixes #1506.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,9 +29,9 @@
   "dependencies": {
     "css-select": "^3.1.2",
     "dom-serializer": "~1.2.0",
+    "domhandler": "^4.0.0",
     "entities": "~2.1.0",
     "htmlparser2": "^6.0.0",
-    "lodash": "^4.17.20",
     "parse5": "^6.0.0",
     "parse5-htmlparser2-tree-adapter": "^6.0.0"
   },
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -29,9 +29,9 @@
   "dependencies": {
     "css-select": "^3.1.2",
     "dom-serializer": "~1.2.0",
+    "domhandler": "^4.0.0",
     "entities": "~2.1.0",
     "htmlparser2": "^6.0.0",
-    "lodash": "^4.17.20",
```
TrySound, line 34: Awesome!

PR #1571, opened 2020-12-20
chore(jsdoc): Base JSDoc on modules, use jsdoc@3.6.6

fb55:
This allows us to drop our fork of JSDoc, which currently relies on a number of insecure packages.


lib/cheerio.js
```diff
@@ -29,11 +29,11 @@ var api = [
...
- * @mixes attributes
- * @mixes css
- * @mixes forms
- * @mixes manipulation
- * @mixes traversing
+ * @mixes module:cheerio/attributes
+ * @mixes module:cheerio/css
+ * @mixes module:cheerio/forms
+ * @mixes module:cheerio/manipulation
+ * @mixes module:cheerio/traversing
```
fb55, line 36: This is the other big change: All of these are now based on modules, removing some previously needed hacks.

PR #1584, opened 2020-12-22
Update TypeScript bindings

f0x52:
to allow setting isDocument to false (and passing null to options)


fb55:
Nice catch, thanks a lot!

Issue #748, opened 2015-09-07
Here is a user-side implementation of .wrapAll()


sfinktah:
I would have written this properly and sent a pull request (using the internal functions for dom manipulation, etc), but the current git version is failing testing.

``` js
var cheerio = require('cheerio');
var _ = require('lodash');

module.exports = (function() {
    var extendCheerio = function extendCheerio($) {
        _.extend($.prototype, {
            wrapAll: function(wrapper) {
                if (this.length < 1) {
                    return this;
                }

                if (this.length < 2 && this.wrap) { // wrap not defined in npm version,
                    return this.wrap(wrapper);      // and git version fails testing.
                }

                var elems = this;
                var section = $(wrapper);
                var marker = $('<div>');
                marker = marker.insertBefore(elems.first()); // in jQuery marker would remain current
                elems.each(function(k, v) {                  // in Cheerio, we update with the output.
                    section.append($(v));
                });
                section.insertBefore(marker); 
                marker.remove();
                return section;                 // This is what jQuery would return, IIRC.
            },
        });
    };

    if ("test") {
        $ = cheerio.load("<html><body><div><p><span>This <em>is <i>test</p><span>More <em>test");
        extendCheerio($);
        $('span').wrapAll('<section>');
        var passed = ($.html() === '<html><body><div><p><section><span>This <em>is <i>test</i></em>'+
                          '</span><span>More <em>test</em></span></section></p></div></body></html>');
        console.log("passed: " + (passed ? "yes" : "no"));
        // console.log($.html() + "\n\n");
    }
    return extendCheerio;
})();
```

ajsnyde:
Doesn't wrap the children, I don't think. I try to $('a').wrapAll('<outer>'):

`<a><b></b><a><a><b></b><a><a><b></b><a>`

and I get `<outer><a></a><a></a></outer><b></b><b></b>`

mhjb:
@sfinktah works for me — thank you very much

fb55:
Happy to accept a PR for this, should someone want to polish things up a bit!

antecuic:
Hi guys!
How did you implement this in code?

mhjb:
@antecuic I did like this: https://github.com/mhjb/studentrights-www-generator/blob/schema/extend_cheerio.js

antecuic:
[...]

PR #1590, opened 2020-12-24
.wrapAll API method

5saviahv:
**.wrapAll** API method

* added tests, what I was able compare with jQuery.
* documentation added based on jQuery
* examples tested with cheerio
* I didn't compare results with implementation in #748


fb55 (approved):
Great works, thanks a lot!

PR #1594, opened 2020-12-25
build(deps): Use cheerio-select@1.0.0

fb55:


bcomnes:
Can we get a release on this?

fb55:
@bcomnes This is only renaming a package, without any functional changes. Any particular reason to get this out more quickly?

bcomnes:
Showing up as top level depreciation warnings. Not the biggest issue but it would be nice to quiet those down.

PR #1595, opened 2020-12-25
child finding fix

5saviahv:
* child finding got broken (it add elements on root element, when it should add them on last branch element)
* updated types (added `wrapAll` function)
* it clones wrapping element taken from DOM (jQuery acts same way)
* trims string white space (preventing first DOM element to be text)

### DIFF
--- a/lib/api/manipulation.js
+++ b/lib/api/manipulation.js
@@ -410,23 +410,27 @@ exports.wrapAll = function (wrapper) {
       wrapper = wrapper.call(this[0]);
     }
 
-    var wrap = this._make(wrapper);
+    // trim whitespace
+    if (typeof wrapper === 'string') {
+      wrapper = wrapper.trim();
+    }
+
+    var wrap = this._make(wrapper).eq(0).clone();
 
     if (this[0].parent) {
       wrap = wrap.insertBefore(this[0]);
     }
 
-    // if html is given as wrapper, wrap may contain text elements
-    var elInsertLocation = { children: wrap };
+    var elInsertLocation = wrap[0];
     var j = 0;
 
     // Find the deepest child. Only consider the first tag child of each node
     // (ignore text); stop if no children are found.
-    while (
-      elInsertLocation &&
-      elInsertLocation.children &&
-      j >= elInsertLocation.children.length
-    ) {
+    while (elInsertLocation && elInsertLocation.children) {
+      if (j >= elInsertLocation.children.length) {
+        break;
+      }
+
       if (elInsertLocation.children[j].type === 'tag') {
         elInsertLocation = elInsertLocation.children[j];
         j = 0;
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -4,6 +4,12 @@ import { ParserOptions } from 'htmlparser2';
 declare namespace cheerio {
   type AttrFunction = (el: Element, i: number, currentValue: string) => any;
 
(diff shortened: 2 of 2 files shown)


lib/api/manipulation.js
```diff
@@ -410,23 +410,27 @@ exports.wrapAll = function (wrapper) {
       wrapper = wrapper.call(this[0]);
     }
 
-    var wrap = this._make(wrapper);
+    // trim whitespace
+    if (typeof wrapper === 'string') {
+      wrapper = wrapper.trim();
```
fb55, line 415: Can you add a test that breaks without this? I'm not sure if it is necessary 🤔

lib/api/manipulation.js
```diff
@@ -410,23 +410,27 @@ exports.wrapAll = function (wrapper) {
...
     var j = 0;
 
     // Find the deepest child. Only consider the first tag child of each node
     // (ignore text); stop if no children are found.
-    while (
-      elInsertLocation &&
-      elInsertLocation.children &&
-      j >= elInsertLocation.children.length
-    ) {
+    while (elInsertLocation && elInsertLocation.children) {
```
fb55, line 429: Whee, forgot to flip the order of the `if` when including it in the loop condition. Could you add a test case that breaks without this change? Also include it in the `while` condition, as it makes the code more readable?

types/index.d.ts
```diff
@@ -4,6 +4,12 @@ import { ParserOptions } from 'htmlparser2';
 declare namespace cheerio {
   type AttrFunction = (el: Element, i: number, currentValue: string) => any;
 
+  /** A Selector is used to select DOM elements from document. */
+  type selectorString = string;
```
fb55, line 8: If we adopt this pattern, we should use it for all places that accept either. I do think it is a nice idea, but perhaps as a separate step?

types/index.d.ts
```diff
@@ -270,7 +280,7 @@ declare namespace cheerio {
       options?: CheerioParserOptions | null,
       isDocument?: boolean
     ): Root;
-    load(element: Element, options?: CheerioParserOptions): Root;
+    load(element: Element | Element[], options?: CheerioParserOptions): Root;
```
fb55, line 283: Great catch!

fb55 (commented):
Nice PR! Left some comments, but this is looking great.

lib/api/manipulation.js
5saviahv: All code here is actually letter by letter from `_wrap` function.

fb55: I saw that — I've included this change in #1596, as tests broke when I migrated to Jest. Looks like tests weren't run properly on Mocha.

types/index.d.ts
5saviahv: True, types `htmlString` and `selectorString` should give you some information what input to expect and `Selector` is already used as interface.

lib/api/manipulation.js
5saviahv: If html string is used, and first characters are any white space character - parser will generate, text node.

It may be overkill, but I managed encounter this during my first tests. I like to test it further, but it should avoid some errors.

lib/api/manipulation.js
5saviahv: Actually tests did fail. In tests I used nested elements and check are elements attached in correct position.

PR #1599, opened 2020-12-26
package.json: fix `format:prettier:raw` script

XhmikosR:
This was failing on Windows native cmd. Using double quotes should work everywhere.


XhmikosR:
BTW another issue I notice on Windows which has `autocrlf` set to `true` by default, is prettier complaining about CRLF. Happy to submit a PR to enforce LF regardless of the git settings.

PR #1600, opened 2020-12-26
Switch to GitHub Actions CI

XhmikosR:
It seems there are a few issues:

1. ~~.travis.yml was using `make travis-test` but that was using `@npm run test:lint` which doesn't exist; I changed it to `@npm run lint` but the makefile isn't up to date.~~ **EDIT**: I switched to `npm test` but IMHO the Makefile should be removed to avoid confusion since it's clearly broken
2. it seems `make travis-test` is failing; `test-cov` also doesn't seem to work. Happy to sort everything out as long as you give me some hints. My suggestion is to move everything to npm scripts only and Actions including coverage.
3. I added another job with `BENCHMARK=true` 
4. We can add further Node.js versions, OS etc. Let me know and I can take a stab at it later. I'd just split lint to a separate action or job and run it only once.

Notes:

1. since Actions don't currently support `LTS` tags, I explicitly set this to 14 in an env var
2. I added npm caching which is what Travis CI was doing by default

EDIT: it seems that `npm test` didn't even run on Travis CI hence why the failures weren't caught sooner.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,5 +45,19 @@ jobs:
       - name: Install npm dependencies
         run: npm ci
 
-      - name: Run tests
-        run: npm run test:jest && npm run test:types
+      - name: Run Jest
+        run: npm run test:jest
+        if: matrix.node != 14
+
+      - name: Run Jest with coverage
+        run: npm run test:jest:cov
+        if: matrix.node == 14
+
+      - name: Test types
+        run: npm run test:types
+
+      - name: Run Coveralls
+        uses: coverallsapp/github-action@master
+        if: matrix.node == 14
+        with:
+          github-token: '${{ secrets.GITHUB_TOKEN }}'
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,67 @@
+name: CI
+
+on:
+  push:
+  pull_request:
+
+env:
+  CI: true
+  FORCE_COLOR: 2
+  NODE: 14.x
+
+jobs:
+  ci:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '${{ env.NODE }}'
+
+      - name: Set up npm cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-v${{ env.NODE }}-${{ hashFiles('package.json') }}-${{ hashFiles('package-lock.json') }}
+          restore-keys: |
+            ${{ runner.OS }}-node-v${{ env.NODE }}-${{ hashFiles('package.json') }}-${{ hashFiles('package-lock.json') }}
+            ${{ runner.OS }}-node-v${{ env.NODE }}-
+
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: Run tests
+        run: npm test
+
+  benchmark:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Clone repository
+        uses: actions/checkout@v2
+
+      - name: Set up Node.js
+        uses: actions/setup-node@v2
+        with:
+          node-version: '${{ env.NODE }}'
+
+      - name: Set up npm cache
+        uses: actions/cache@v2
+        with:
+          path: ~/.npm
+          key: ${{ runner.os }}-node-v${{ env.NODE }}-${{ hashFiles('package.json') }}-${{ hashFiles('package-lock.json') }}
+          restore-keys: |
+            ${{ runner.OS }}-node-v${{ env.NODE }}-${{ hashFiles('package.json') }}-${{ hashFiles('package-lock.json') }}
+            ${{ runner.OS }}-node-v${{ env.NODE }}-
+
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: Run benchmarks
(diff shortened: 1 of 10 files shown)


XhmikosR:
I decided to split the actions so that lint and tests are done faster. So, things that remain:

1. should we get rid of Makefile completely? I vote for yes since it's outdated and confusing to keep the build scripts in two places
2. should I add more Node.js versions in tests? If so, which ones?
3. should I add more OS'es in tests? If so, which ones and in which tests?

fb55:
Tests should be fixed with https://github.com/cheeriojs/cheerio/pull/1596

XhmikosR:
@fb55 so, how do you think we should proceed with the remaining TODO? Also, is there something you want me to change?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
+name: CI
+
+on:
+  push:
+    branches-ignore:
+      - 'dependabot/**'
```
fb55, line 6: Will we still run tests for Dependabot PRs with this?

.github/workflows/ci.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Install npm dependencies
+        run: npm ci
+
+      - name: Run Jest
+        run: npm run test:jest
+        if: matrix.node != 14
+
+      - name: Run Jest with coverage
+        run: npm run test:jest:cov
+        if: matrix.node == 14
```
fb55, line 54: Would it make sense to move this to an env var, so that there is only a single reference to the Node version used for coverage?

.github/workflows/site.yml
```diff
@@ -12,7 +17,15 @@ jobs:
       - uses: actions/checkout@v2
       - uses: actions/setup-node@v2
         with:
-          node-version: 14.x
+          node-version: '${{ env.NODE }}'
```
fb55, line 20: This is great 👌

Readme.md
```diff
@@ -3,8 +3,8 @@
 <h5 align="center">Fast, flexible & lean implementation of core jQuery designed specifically for the server.</h5>
 
 <div align="center">
-  <a href="https://travis-ci.org/cheeriojs/cheerio">
-    <img src="https://img.shields.io/travis/cheeriojs/cheerio/main" alt="Travis CI">
+  <a href="https://github.com/cheeriojs/cheerio/actions?query=workflow%3ACI+branch%3Amain">
+    <img src="https://img.shields.io/github/workflow/status/cheeriojs/cheerio/CI/main" alt="Build Status">
```
fb55, line 7: Should lint and CI be in the same file (as different jobs), to make this cover both? (That's the current behavior, but I can also see a point about having them be separate.)

fb55 (commented):
This looks great. Added some questions.

.github/workflows/ci.yml
XhmikosR: Yup, since depndabot always opens a PR, this just skips the branches.

.github/workflows/ci.yml
XhmikosR: Can do, how do you want to name the env var? Also, do you want the env var to be in the global scope or under the job?

Readme.md
XhmikosR: The reason I split the actions is so that everything is faster since they run in parallel this way. I doubt having a badge to lint is something that people should care about :)

.github/workflows/ci.yml
fb55: The pattern in the site action is great, maybe just copy that?

Readme.md
fb55: Wouldn't separate jobs be executed in parallel as well?

.github/workflows/ci.yml
XhmikosR: OK, it's just that it might not be self-explanatory using `NODE`, but I can add a comment.

Also, we could just run jest with `--coverage` regardless of the Node version, and just run coveralls only on Node.js 14. That would be less lines.

Readme.md
XhmikosR: Yeah, but I personally find this cleaner. Each workflow does one thing. That being said, the types test could probably be moved to the lint workflow.

.github/workflows/ci.yml
fb55: Saw you changed the name to NODE_COV, which should be easy enough to distinguish.

> we could just run jest with --coverage regardless of the Node version

I like that idea as well. But the current slight duplication isn't too bad either 🙂

Readme.md
fb55: Makes sense & I can live with this.

.github/workflows/ci.yml
XhmikosR: Your call. There isn't so much of a difference, just 3 seconds. I'll leave it as is and you can change it later if needed.

fb55:
I'm happy to go forward with this. Super excited to see this land.

@XhmikosR The PR is still a draft. Let me know when you feel ready to proceed.

XhmikosR:
I'm still not sure about keeping the Makefile around. It's unnecessary duplication and it seems Makefile has a script that is not present in package.json (xyz).

I'll leave that to you to sort later :)

PS. I'm still being bitten by #1602 since my dev VM is on Windows.

fb55:
Good point about the Makefile. Thanks so much for this PR, this moves us forward considerably.

XhmikosR:
@fb55 a few more things to do:

1. go to the repo apps/integrations/hooks and remove any Travis CI related stuff
2. adapt any branch protection
3. go on coveralls and change the threshold for failure to something like 0.5; by default, even a .001 decrease will mark the build as failed

PR #1601, opened 2020-12-26
Add CodeCQL Action

XhmikosR:
Runs only on main and PRs against main, and on schedule.

* https://github.com/github/codeql-action


fb55:
Thanks!

PR #1603, opened 2020-12-27
Fix a few redirects

XhmikosR:


Readme.md
```diff
@@ -6,7 +6,7 @@
   <a href="https://travis-ci.org/cheeriojs/cheerio">
     <img src="https://travis-ci.org/cheeriojs/cheerio.svg?branch=main" alt="Travis CI" />
   </a>
-  <a href="https://coveralls.io/r/cheeriojs/cheerio">
+  <a href="https://coveralls.io/github/cheeriojs/cheerio">
     <img src="https://img.shields.io/coveralls/cheeriojs/cheerio.svg?branch=main&style=flat" alt="Coverage" />
```
XhmikosR, line 10: I was thinking, might be worth switching to shields.io for all badges; less domains, less DNS lookups and consistency will be enforced. What do you think?

fb55: I like it! The repo isn't an issue as Github proxies all images anyway, but this might be a nice thing to do for the website.

fb55:
Great fix, thanks!

PR #1604, opened 2020-12-27
codeql.yml: fix formatting

XhmikosR:
It doesn't seem Travis CI is running `npm test`, I think you should move with #1600 sooner than later :)


fb55:
Thanks! Already solved in 71d2aaf9174756355d9d9424fa61685662fbe672 😄

PR #1605, opened 2020-12-27
dependabot: update GitHub Actions too

XhmikosR:
I went with a weekly schedule on Wednesday, but feel free to let me know or edit the PR to adapt it to your needs.

BTW, have you thought about using `versioning-strategy: increase` for the npm check?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -5,3 +5,9 @@ updates:
     schedule:
       interval: daily
     open-pull-requests-limit: 10
+  - package-ecosystem: 'github-actions'
+    directory: '/'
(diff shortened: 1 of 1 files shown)


fb55:
Awesome, thanks a lot!

> BTW, have you thought about using `versioning-strategy: increase` for the npm check?

I haven't, but reading about it, it does make a lot of sense to enable it.

fb55:
Did so in `71d2aaf9174756355d9d9424fa61685662fbe672`

XhmikosR:
Also, you might want to consider switching to weekly for the npm updates. Daily is a little too much :)

PR #1610, opened 2020-12-27
JSDocs: enable Markdown IDs

XhmikosR:
**Untested yet**

I need to check if there are any duplicate IDs and how the plugin handles that case.


XhmikosR:
Should fix the broken anchors on the website like the sponsors badge :)

fb55:
Very nice, thanks!

PR #1611, opened 2020-12-27
Update Readme.md badges

XhmikosR:
~~I removed the coveralls badge since the integration doesn't work right now.~~

Restored this since it seems to work fine in #1600


fb55:
Thanks!

PR #1612, opened 2020-12-27
eslint: remove `--ignore-path`

XhmikosR:
ESLint respects .gitignore


fb55:
Nice find! Looks like we can get rid of the `.prettierignore` all together, as it now covers a subset of the `.gitignore`.

XhmikosR:
Indeed, but I thought I'd leave that for the maintainers. :)

fb55:
😄 Sounds good, will follow up then. Thanks for digging this deep into the project, this has been great!

PR #1613, opened 2020-12-27
Remove the unused direct entities package

XhmikosR:
This seems redundant as a direct dependency AFAICT.


fb55:
Good catch, this was never removed after dom-serializer was spun out.

PR #1619, opened 2020-12-28
Update dependencies and regenerate package-lock.json

XhmikosR:
So that everything is in sync after the dependabot config change.


fb55:
Nice, thanks!

Issue #1045, opened 2017-06-23
Stringify behaviour differs between cheerio and jQuery


stevenvachon:
jQuery fragment `html()`:
```js
$("<html></html>").html()
//-> ""
```
cheerio fragment `html()`:
```js
const $ = cheerio.load("")

$("<html></html>").html()
//-> <html><head></head><body></body></html>
```

---

jQuery `toString()`:
```js
$(":root").toString()
//-> [object Object]
```
cheerio `toString()`:
```js
const $ = cheerio.load("")

$(":root").toString()
//-> <html><head></head><body></body></html>
```

@inikulin @fb55

jcrben:
@fb55 @inikulin this could use the label jQuery Differences https://github.com/cheeriojs/cheerio/issues?q=is%3Aopen+is%3Aissue+label%3A%22jQuery+Differences%22

fb55:
Thanks @jcrben!

fb55:
I believe Cheerio's behaviour is more useful in a server context, so closing this as a won't fix.

Issue #1083, opened 2017-09-19
script.html(text) will parse `<` in the text


mingelz:
cheerio version: 1.0.0-rc.2

```javascript
var script = $('script').eq(0)

// correct:
script.html('a < b')
script.html() // a < b

// fault: (if remove spaces)
script.html('a<b')
script.html() // a

// correct:
script.replaceWith('<script>a<b</script>')
```

fb55:
I assume this is some issue with the context not being passed when parsing fragments. cc @inikulin

5saviahv:
I look into

PR #1622, opened 2020-12-29
Simple parse for single elements

5saviahv:
It parses single `<x></x>, <x/>, <x>` style elements.
Currently `cheerio('<html></html>')` gives you empty result.

jQuery acts same way and may help solving #1045

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -5,6 +5,7 @@ var DomUtils = require('htmlparser2').DomUtils;
 var parseWithHtmlparser2 = require('./parsers/htmlparser2').parse;
 var parseWithParse5 = require('./parsers/parse5').parse;
 var Document = require('domhandler').Document;
+var Element = require('domhandler').Element;
 
 /*
   Parser
@@ -12,6 +13,10 @@ var Document = require('domhandler').Document;
 exports = module.exports = function parse(content, options, isDocument) {
   if (typeof Buffer !== 'undefined' && Buffer.isBuffer(content)) {
     content = content.toString();
+  }
+
+  if (!isDocument) {
+    content = _parseSimple(content);
   }
 
   if (typeof content === 'string') {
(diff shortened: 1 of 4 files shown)


fb55:
Thanks for looking into this! I think the proper solution here is to pass the fragment context to `parseFragment`; see https://github.com/inikulin/parse5/blob/master/packages/parse5/docs/index.md

I am not sure what the context should be for `$('<a>')`-style calls (might be the root, might be eg. the body tag). That would take some reading of jQuery's source code to figure out.

5saviahv:
I have poking around in jQuery source code and they use exact same method for single html tags and uses parseHTML if anything else is given.

I am not sure is it there only for performance or maybe it is there for plugins, but with parseFragment you are not able parse some basic body tags like `html, body, head`.

You may need those body tags for building page or parts of the page. for work around you have to use `cheerio.load()` or use some random tag name like `<mybody>` and later rename it.

5saviahv:
I still think it is correct since jQuery uses same method in [v1.12](https://github.com/jquery/jquery/blob/e09907ce152fb6ef7537a3733b1d65ead8ee6303/src/core/parseHTML.js#L27) / [v2.2](https://github.com/jquery/jquery/blob/b14ce54334a568eaaa107be4c441660a57c3db24/src/core/parseHTML.js#L27) / [v3.x](https://github.com/jquery/jquery/blob/bf06dd477a1bf829c68d64c7de9e2262ac8973bb/src/core/parseHTML.js#L51) / [master](https://github.com/jquery/jquery/blob/8969732518470a7f8e654d5bc5be0b0076cb0b87/src/core/parseHTML.js#L40)

I tried find the way with parse5 Fragment context, only way I managed get even some results was If context is full documents with all elements, but all fragments becomes then full html documents.

fb55:
Thanks for the links @5saviahv, super helpful. I am now convinced that this is the right move forward. Could you add a test case from #1045 to make sure everything continues to work?

fb55:
Adding the context should solve #1083 by the way, which would be a cool follow-up if you already have the code 🧐

5saviahv:
It fails because domhandlers Element does not have namespace parameter specified.

PR #1625, opened 2021-01-01
Change how html string will be treated

5saviahv:
When jQuery tests html string they later use that matched string as html string, but cheerio discards it and uses original.
``` javascript
jQuery('\n<div></div>\n')
//=> [div]
cheerio('\n<div></div>\n')
//=> [text,div,text]
```
[L39](https://github.com/jquery/jquery/blob/8969732518470a7f8e654d5bc5be0b0076cb0b87/src/core/init.js#L39)

### DIFF
--- a/lib/cheerio.js
+++ b/lib/cheerio.js
@@ -5,7 +5,7 @@
 var parse = require('./parse');
 var defaultOptions = require('./options').default;
 var flattenOptions = require('./options').flatten;
-var isHtml = require('./utils').isHtml;
+var getHtml = require('./utils').getHtml;
 
 /*
  * The API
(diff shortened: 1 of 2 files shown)


fb55:
Very interesting PR. I know this is similar to how jQuery works, but want to do some more testing before going forward. (Hopefully in the next few days.)

One thing to look into: Afaict, right now, eg. `$('<p>foo<p>bar')` would skip the text of the second `<p>`. Not sure that this is the same for jQuery.

5saviahv:
Actually it is same in jQuery, but I think it is wrong.

5saviahv:
probably not worth the hassle

fb55:
This was a great find @5saviahv, and a very tough call. IMHO that this is an area where (1) jQuery's behavior is surprising and doesn't seem to necessarily match user intent and (2) making a breaking change that doesn't align with jQuery doesn't really seem worth it.

Thanks a lot for opening this independently of the outcome!

PR #1627, opened 2021-01-01
.html() send context to parse5

5saviahv:
using proper context when parsing in .html() 

It solves #1083

### DIFF
--- a/lib/api/manipulation.js
+++ b/lib/api/manipulation.js
@@ -753,12 +753,14 @@ exports.html = function (str) {
     return html(this[0].children, this.options);
   }
 
-  var opts = this.options;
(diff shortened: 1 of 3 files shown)


fb55:
Great job @5saviahv!

bodinsamuel:
Hello, this seems to break when used in combinaison with _useHtmlParser2
I know htmlParser2 is somewhat hidden so, is it fixable or is it expected/won't fix? ☺️ 

https://runkit.com/embed/kv9cok8uese9

5saviahv:
hmm, it seems to be my mistake/typo

bodinsamuel:
Thanks @5saviahv 🙏🏻

PR #1629, opened 2021-01-02
package.json: Add a "bench" alias script

XhmikosR:


fb55:
I'm not a super big fan of having aliases in the `package.json`. Should we just rename the original script?

XhmikosR:
Your call, but alias help because they work in more cases. Feel free to close the PR, it's just I kept running `npm run bench`. This should allow both cases to work.

fb55:
Decided to merge this, as we don't have insane amounts of scripts at the moment.

fb55:
Thanks @XhmikosR!

PR #1633, opened 2021-01-02
single element parser

5saviahv:
re done #1622 with parse5 treeadapter.

### DIFF
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -5,6 +5,7 @@
 var DomUtils = require('htmlparser2').DomUtils;
 var parseWithHtmlparser2 = require('./parsers/htmlparser2').parse;
 var parseWithParse5 = require('./parsers/parse5').parse;
+var parseElement = require('./parsers/parse5').parseElement;
 var Document = require('domhandler').Document;
 
 /*
@@ -13,6 +14,10 @@ var Document = require('domhandler').Document;
 exports = module.exports = function parse(content, options, isDocument) {
   if (typeof Buffer !== 'undefined' && Buffer.isBuffer(content)) {
     content = content.toString();
+  }
+
+  if (!isDocument) {
+    content = parseElement(content);
   }
 
   if (typeof content === 'string') {
--- a/lib/parse.js
+++ b/lib/parse.js
@@ -5,7 +5,6 @@
 var DomUtils = require('htmlparser2').DomUtils;
 var parseWithHtmlparser2 = require('./parsers/htmlparser2').parse;
 var parseWithParse5 = require('./parsers/parse5').parse;
-var parseElement = require('./parsers/parse5').parseElement;
 var Document = require('domhandler').Document;
 
 /*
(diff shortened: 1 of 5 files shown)


lib/parsers/parse5.js
```diff
@@ -30,3 +30,22 @@ exports.render = function (dom) {
     { treeAdapter: htmlparser2Adapter }
   );
 };
+
+// Attempts to match single html elements
+// like <..></..>, <../>, <..>
+var rsingleTag = /^<([a-z][^/\0>:\x20\t\r\n\f]*)[\x20\t\r\n\f]*\/?>(?:<\/\1>)?$/i;
+// HTML namespace
+var ns_html = 'http://www.w3.org/1999/xhtml';
```
XhmikosR, line 38: Shouldn't this be `nsHtml`? Or even better drop the var since it's used once?

lib/parsers/parse5.js
```diff
@@ -30,3 +30,22 @@ exports.render = function (dom) {
     { treeAdapter: htmlparser2Adapter }
   );
 };
+
+// Attempts to match single html elements
+// like <..></..>, <../>, <..>
+var rsingleTag = /^<([a-z][^/\0>:\x20\t\r\n\f]*)[\x20\t\r\n\f]*\/?>(?:<\/\1>)?$/i;
```
XhmikosR, line 36: This could probably be optimized to this (untested):

```js
var rsingleTag = /^<([a-z][^/\0>: \t\r\n\f]*)[ \t\r\n\f]*\/?>(?:<\/\1>)?$/i;
```

5saviahv: true \x20 is single space

test/cheerio.js
```diff
@@ -44,6 +44,13 @@ describe('cheerio', function () {
     expect($h2[0].tagName).toBe('h2');
   });
 
+  it('should be able to create html without a root or context, html tag', function () {
+    var $tag = cheerio('<html>');
+    expect($tag).not.toHaveLength(0);
```
XhmikosR, line 49: This is a little moot since you check for length 1 in the next line :)

5saviahv: haha, I simply copied test above. I looked it, but since it was there I kept it.

lib/parsers/parse5.js
5saviahv: I put it in var because prettier formatting in sentence looked awkward, but naming you are right.

lib/parse.js
```diff
@@ -14,6 +15,10 @@ exports = module.exports = function parse(content, options, isDocument) {
     content = content.toString();
   }
 
+  if (!isDocument) {
+    content = parseElement(content);
```
fb55, line 19: This will always parse content using `parse5` and won't respect the `_useHtmlParser2` or `xmlMode` options (see right below).

One alternative implementation would be to always use `new Element` here, and initialize the parse5 extensions to empty objects:

```js
const elem = new Element(THE_TAG_NAME, {})
elem[ "x-attribsNamespace"] = {}
elem["x-attribsPrefix"] = {};
```

5saviahv: True with xml it would generate element with incorrect namespace and htmlparser2 in general dont need it.

Now it will be co-parser only with parse5.

Problem with `new Element()` is parse5 expects every elements to have namespace and may fail to recognise them as html if it is missing.

PR #1635, opened 2021-01-02
attributes: remove unneeded escapes

XhmikosR:


lib/api/attributes.js
```diff
@@ -23,7 +23,7 @@ var primitives = {
 // Attributes that are booleans
 var rboolean = /^(?:autofocus|autoplay|async|checked|controls|defer|disabled|hidden|loop|multiple|open|readonly|required|scoped|selected)$/i;
 // Matches strings that look like JSON objects or arrays
-var rbrace = /^(?:\{[\w\W]*\}|\[[\w\W]*\])$/;
+var rbrace = /^(?:{[\w\W]*}|\[[\w\W]*])$/;
```
fb55, line 26: To simplify further, this should be equivalent:

```suggestion
var rbrace = /^(?:{[^]*}|\[[^]*])$/;
```

XhmikosR: I just split the patch from my xo branch, but feel free to tweak it later. I can't check if your suggestion is fine right now :)

fb55:
Thanks!

PR #1637, opened 2021-01-02
parent() may fail in functions

5saviahv:
parent() don't work well in functions if selector is undefined

``` javascript
function getParent(el, parent) {
    return el.parent(parent).not('body');
}

var parent = getParent($('#rating span'))

```
In [.parent()](https://github.com/cheeriojs/cheerio/blob/703ec166d8b98aab714ad8930858bbefaed847b1/lib/api/traversing.js#L89) has condition:
``` javascript
...
  if (arguments.length) {
...
```
but it will fail if used like in function above, since `arguments.length` will be 1 even when `selector` is actually undefined.

### DIFF
--- a/lib/api/traversing.js
+++ b/lib/api/traversing.js
@@ -86,7 +86,7 @@ exports.parent = function (selector) {
     }
   });
 
-  if (arguments.length) {
+  if (selector) {
     set = exports.filter.call(set, selector, this);
   }
 
--- a/test/api/traversing.js
+++ b/test/api/traversing.js
@@ -619,6 +619,12 @@ describe('$(...)', function () {
       expect(result[1].attribs.id).toBe('vegetables');
     });
 
+    it('(undefined) : should not throw an exception', function () {
+      expect(function () {
+        $('li').parent(undefined);
+      }).not.toThrow();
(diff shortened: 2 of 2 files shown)


XhmikosR:
This reminds me, shouldn't everything run in strict mode?

fb55:
Nice find! Happy to merge this, just out of curiosity: Is this actually a fix, or just a short-circuit for an empty input?

5saviahv:
What do you mean ? `arguments` variable is array like, I would call `arguments.length` a hack.

if variable parent is not defined it gets value **undefined**, but when I use it in function like:
* `.parent()`, variable **arguments** will be `[]` so length is 0
* `.parent(parent)`, variable **arguments** will be `[undefined]` and length is 1

variable **selector** will be **undefined** both cases

fb55:
What I was trying to get at was _does this actually produce an error?_ And yes, it does:

```js
$('<a>').parent(undefined)
// => Uncaught TypeError: Cannot read property 'cheerio' of undefined
```

Could you add a test case that covers this?

fb55:
Awesome, thanks!

PR #1638, opened 2021-01-02
ESLint: enable eqeqeq rule expect for null

XhmikosR:
This should match what's currently used AFAICT. I left `null` out because it's handy to keep it.


fb55:
Very nice, happy to go forward with this!

PR #1639, opened 2021-01-03
Prefer Jest's toBe true/false matcher.

XhmikosR:
Also, remove some trailing spaces from descriptions.

Might be worth adding https://github.com/jest-community/eslint-plugin-jest later :)

### DIFF
--- a/test/api/attributes.js
+++ b/test/api/attributes.js
@@ -283,7 +283,7 @@ describe('$(...)', function () {
       expect(Object.keys(data)).toHaveLength(0);
     });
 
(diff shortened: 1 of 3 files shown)


5saviahv:
Interesting [toBe(true) vs toBeTruthy() vs toBeTrue()](https://stackoverflow.com/questions/32615713/tobetrue-vs-tobetruthy-vs-tobetrue)

fb55:
Nice one 🙂

> Might be worth adding https://github.com/jest-community/eslint-plugin-jest later :)

Good call, looking into it!

PR #1640, opened 2021-01-04
added some tests

5saviahv:
added/changed some tests so more conditions would be covered

### DIFF
--- a/lib/api/traversing.js
+++ b/lib/api/traversing.js
@@ -697,7 +697,7 @@ exports.not = function (match, container) {
     });
   }
 
-  return container._make(elements);
+  return this._make(elements);
 };
 
 /**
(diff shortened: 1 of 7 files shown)


lib/api/traversing.js
```diff
@@ -697,7 +697,7 @@ exports.not = function (match, container) {
     });
   }
 
-  return container._make(elements);
```
fb55, line 700: This is using the same pattern as `filter` right now. `filter` can be called with `this` being an array. We don't do that anywhere for `not` afaict, but maintaining the same pattern should be good in any case.

fb55:
This looks great! The change in `.not` reduces consistency and should probably be reverted, otherwise these are some great changes.

lib/api/traversing.js
5saviahv: ok, reverted

fb55:
Thanks for following up, and nice patch!

PR #1642, opened 2021-01-04
chore(lint): Add jest & node eslint plugins

fb55:
And fixes all reported issues.


benchmark/benchmark.js
```diff
@@ -1,13 +1,10 @@
-#!/usr/bin/env node
```
XhmikosR, line 1: I'd keep the shebang since the script is a CLI script

fb55: `eslint-plugin-node`'s recommended config only allows a shebang when the script is actually a part of the `package.json`'s `bin`s. There isn't too much of a point for exposing this externally, so removing it seems reasonable. We always call this with `node` anyway, so nothing should change.

XhmikosR: Yeah, exactly why the shebang should stay :)

IMO this is wrong, and the file should be treated as a CLI script since that's what it is.

XhmikosR: See also https://github.com/mysticatea/eslint-plugin-node/issues/96

and https://github.com/xojs/xo/blob/master/config/plugins.js#L309-L310

Personally, I'd disable the rule and keep the shebang.

.eslintrc.json
```diff
@@ -3,9 +3,18 @@
...
+  "extends": [
+    "eslint:recommended",
+    "plugin:jsdoc/recommended",
+    "plugin:jest/recommended",
+    "plugin:node/recommended",
+    "prettier"
+  ],
   "globals": { "Set": true, "Symbol": true },
   "rules": {
+    // TODO
```
XhmikosR, line 15: Might be worth renaming .eslintrc.json to .eslintrc.js later so that the file is valid, but that's nitpicking since ESLint works with comments in JSON too.

BTW you could set it to 4.5.0 so that it works, but the discrepancy between the engines version and the current code will still exist.

benchmark/benchmark.js
fb55: Thanks a lot for the references. This was super helpful, I'm convinced 😄

.eslintrc.json
fb55: Yeah — we really need to find a solution for this. I've disabled the rule for now and will revisit it soon.

PR #1648, opened 2021-01-04
jsdocs: switch to clean-jsdoc-theme

XhmikosR:
Not tied to this, just wanted something better looking. If anyone has any theme suggestions let me know :)

* Theme options: <https://github.com/ankitskvmdam/clean-jsdoc-theme>
* Preview: <https://friendly-poincare-2097a2.netlify.app>

PS. Not sure if it's worth adding Netlify to the repo for automatic PR previews, but the free plan should suffice. Another alternative would be Vercel.


fb55:
Thanks a lot for this! I'm not the biggest fan of the `clean-jsdoc-theme`, but I do agree that it is a step forward for our website.

I have tried several JSDoc themes, but none of them came close to what I'd like to see. TypeDoc does seem to have some nicer themes, making me consider #1641 even more 😄

Issue #850, opened 2016-04-19
Add .unwrap() method


asumaran:

bertyhell:
Should be pretty easy to implement:
`$textbox.replaceWith($textbox.html());`

PR #1651, opened 2021-01-04
Added unwrap function

5saviahv:
Should solve #850

### DIFF
--- a/lib/api/manipulation.js
+++ b/lib/api/manipulation.js
@@ -390,7 +390,7 @@ exports.wrapInner = _wrap(function (el, elInsertLocation, wrapperDom) {
  *   //     <p>World</p>
  *   //   </div>
  *
- * @param {string} selector - A selector to check the parent element against. If
+ * @param {string} [selector] - A selector to check the parent element against. If
  *     an element's parent does not match the selector, the element won't be unwrapped.
  * @see {@link https://api.jquery.com/unwrap/}
  * @returns {Cheerio} The instance itself, for chaining.
--- a/lib/api/manipulation.js
+++ b/lib/api/manipulation.js
@@ -370,7 +370,7 @@ exports.wrapInner = _wrap(function (el, elInsertLocation, wrapperDom) {
  *
  * @example <caption>without selector</caption>
  *   const $ = cheerio.load(
- *     '<div id=test>\n  <p>Hello</p></div>\n  <div><p>World</p>\n</div>'
+ *     '<div id=test>\n  <div><p>Hello</p></div>\n  <div><p>World</p></div>\n</div>'
  *   );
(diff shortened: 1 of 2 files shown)


lib/api/manipulation.js
```diff
@@ -363,6 +363,47 @@ exports.wrapInner = _wrap(function (el, elInsertLocation, wrapperDom) {
   updateDOM(wrapperDom, el);
 });
 
+/**
+ * The .unwrap() function, removes the parents of the set of matched elements
+ * from the DOM, leaving the matched elements in their place.
+ *
+ * @example <caption>without selector</caption>
+ *   const $ = cheerio.load(
+ *     '<div id=test>\n  <p>Hello</p></div>\n  <div><p>World</p>\n</div>'
```
fb55, line 372: This example doesn't seem entirely correct. Why wouldn't `#test` be deleted?

lib/api/manipulation.js
```diff
@@ -363,6 +363,47 @@ exports.wrapInner = _wrap(function (el, elInsertLocation, wrapperDom) {
...
+ *     '<div id=test>\n  <p>Hello</p>\n  <b><p>World</p></b>\n</div>'
+ *   );
+ *   $('#test p').unwrap('b');
+ *
+ *   //=> <div id=test>
+ *   //     <p>Hello</p>
+ *   //     <p>World</p>
+ *   //   </div>
+ *
+ * @param {string} selector - A selector to check the parent element against. If
```
fb55, line 392: Mark this as optional?

```suggestion
 * @param {string} [selector] - A selector to check the parent element against. If
```

lib/api/manipulation.js
```diff
@@ -363,6 +363,47 @@ exports.wrapInner = _wrap(function (el, elInsertLocation, wrapperDom) {
...
+ *     an element's parent does not match the selector, the element won't be unwrapped.
+ * @see {@link https://api.jquery.com/unwrap/}
+ * @returns {Cheerio} The instance itself, for chaining.
+ */
+exports.unwrap = function (selector) {
+  var self = this;
+  this.parent(selector)
```
fb55, line 399: I see where the motivation for the parent fix came from 😄

fb55 (commented):
Thanks for looking into this! Two minor issues, otherwise this should be good to go.

lib/api/manipulation.js
5saviahv: true, small but nasty bug

lib/api/manipulation.js
5saviahv: You are correct and I updated example, sorry. Second one is correct.

fb55:
Awesome, thanks a lot!

PR #1653, opened 2021-01-05
added some tests

5saviahv:
* eslint ignore jest/expect-expect
* I tried add some diversity with tests

### DIFF
--- a/test/api/attributes.js
+++ b/test/api/attributes.js
@@ -175,8 +175,9 @@ describe('$(...)', function () {
     });
 
     it('(invalid key) : invalid prop should get undefined', function () {
-      var attr = checkbox.prop('lol');
-      expect(attr).toBe(undefined);
+      expect(checkbox.prop('lol')).toBe(undefined);
+      expect(checkbox.prop(4)).toBe(undefined);
+      expect(checkbox.prop(true)).toBe(undefined);
     });
 
     it('(key, value) : should set prop', function () {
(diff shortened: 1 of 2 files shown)


test/cheerio.js
```diff
@@ -369,4 +376,33 @@ describe('cheerio', function () {
...
+      expect(utils.cssCase('neverSayNever')).toBe('never-say-never');
+      expect(utils.cssCase('CSSCase')).toBe('-c-s-s-case');
+    });
+
+    it('cloneDom : should be able clone single Elements', function () {
+      var result = [];
+      utils.domEach(cheerio('<p>A</p><p>B</p>'), function (i, el) {
+        result = result.concat(utils.cloneDom(el));
+      });
+      expect(result).toHaveLength(2);
```
fb55, line 399: Can we check the actual shape of the response here?

test/api/attributes.js
```diff
@@ -359,7 +370,7 @@ describe('$(...)', function () {
         balls: 'giandor',
       });
       // Adding as string.
-      var b = $('.linth').data('snack', 'chocoletti');
+      var b = $('.linth').data('snack', 'chocoletti').data(4, 'chocoletti');
```
fb55, line 373: Check the value of `data(4)`?

fb55 (commented):
Nice additions! Two comments, otherwise this is great.

test/api/attributes.js
5saviahv: yes I used not a string for test. and it doesnt break.

It managed to throw when I did same with setAttr.

5saviahv: I added throw check

test/cheerio.js
5saviahv: I wanted test it with toEqual but because circular reference it will hang jest, so I removed it. But I added some tests.

fb55:
Great PR 👏

PR #1654, opened 2021-01-05
Update package.json

pijalu:
Fix cheerio-select deps: Do not use latest cheerio-select as it breaks build on older installs

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,7 +18,7 @@
     "node": ">= 0.6"
   },
   "dependencies": {
-    "cheerio-select": "*",
+    "cheerio-select": "0.0.3",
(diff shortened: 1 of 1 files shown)


pijalu:
Sadly I don't seems to be able to create a PR on cheerio:0.12.4 (only tags) where this fix is required...
Or I am missing something ?

pijalu:
https://github.com/cheeriojs/cheerio/compare/0.12.4...pijalu:0.12.4 does not allow to open PR on TAGS on your side AFAIK... so I can't submit better than "against" main...

5saviahv:
Yes It needs branch

PR #1656, opened 2021-01-06
Fix .add

### LINKED ISSUE #834: Weird add() behavior modifies the selection when changing child attributes
Hello, 

We've been using the cheeriojs library with great success but have run into a strange behavior represented by this test case:

```js
    it ('modifying nested selections should not impact the parent', function() {
      var apple = $('.apple'),
          pear = $('.pear');

      var apple_pear = apple.add(pear);

      // applies red to apple and pear
      apple_pear.addClass('red');

      expect(apple.hasClass('red')).to.be.ok(); // this is true
      expect(pear.hasClass('red')).to.be.ok(); // this is true

      // applies green to pear... AND apple
      pear.addClass('green');
      expect(pear.hasClass('green')).to.be.ok(); //currently this is true
      expect(apple.hasClass('green')).to.be.ok(); // and this is true!
    });

=> <ul id="fruits">
   <li class="apple red green">Apple</li>
   <li class="orange">Orange</li>
   <li class="pear red green">Pear</li>
</ul>
```

Inside exports.add(), modifying the code to create a clone of  the "other" selection seemed to work (any further modifications to pear applied to pear as intended), but this seemed to break  the uniqSort functionality, so the result was not sorted as expected and adding a class to the new selection still only modified the first object in the selection. Besides breaking many things since other is not always clone-able)

Passing a clone'd pear to the add method fixes this odd behavior (child elements can then be modified without modifying the parent), but no changes to the apple_pear selection are applied to the pear elements:

```js
apple.add(pear.clone()).addClass('red');     
$.html()
=>
<ul id="fruits"><li class="apple red">Apple</li><li class="orange">Orange</li><li class="pear">Pear</li></ul>

pear.addClass('green');
$.html()
=>
<ul id="fruits"><li class="apple red">Apple</li><li class="orange">Orange</li><li class="pear green">Pear</li></ul>
```

Curious if anyone has run into this before.

In nontrivial examples, this has a large impact on the way we modify nested table and list elements.

Reproduced on Node 0.12.7 and on 4.4.0

5saviahv:
if **other** is **cheerio object**, `_make` will return it as unchanged. So if `add` changes something there it also affects original **other** object.

Should fix #834

### DIFF
--- a/lib/api/traversing.js
+++ b/lib/api/traversing.js
@@ -902,14 +902,8 @@ exports.end = function () {
  */
 exports.add = function (other, context) {
   var selection = this._make(other, context);
-  var contents = uniqueSort(selection.get().concat(this.get()));
-
-  for (var i = 0; i < contents.length; ++i) {
-    selection[i] = contents[i];
-  }
-  selection.length = contents.length;
-
-  return selection;
+  var contents = uniqueSort(this.get().concat(selection.get()));
+  return this._make(contents);
 };
 
 /**
--- a/test/api/traversing.js
+++ b/test/api/traversing.js
@@ -1383,6 +1383,21 @@ describe('$(...)', function () {
           expect($selection[3]).toBe($pear[0]);
         });
       });
+
+      it('modifying nested selections should not impact the parent [#834]', function () {
+        var apple_pear = $apple.add($pear);
+
+        // applies red to apple and pear
+        apple_pear.addClass('red');
+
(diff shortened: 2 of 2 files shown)


fb55:
Nice one! I am a bit surprised that this wasn't implemented this way before, as `prevObject` wasn't updated before.

Also, we should rename `_make` to `pushStack`, to match jQuery — I've opened #1665 to track.

PR #1659, opened 2021-01-07
tests: prefer using `.toBeUndefined()`

XhmikosR:
There's one case where tests fail; I added a comment there, but I'm not sure if it's a bug or not.

EDIT: I think it's intended but please confirm so that I revert that part.


fb55:
The string value is intended, but wrong. jQuery does stringily values here, but `undefined` as a value is still removed.

Let's keep the test intact here, and track this in a separate issue?

5saviahv:
maybe add TODO comment for a test

fb55:
Great patch, thanks!

PR #1668, opened 2021-01-11
fix jsdoc so wrap & wrapInner would be shown as methods

5saviahv:
fix **jsdoc** so **wrap** & **wrapInner** would be shown as methods


fb55:
Thanks!

PR #1669, opened 2021-01-11
fix(typing): AttrFunction

maxma241:
code snippet
https://github.com/cheeriojs/cheerio/blob/main/lib/api/attributes.js#L113
```js
value.call(el, i, el.attribs[name])
```

the first argument is thisArg not the function args


fb55:
Nice find, thanks!

PR #1673, opened 2021-01-13
Update CI config

XhmikosR:
* remove unneeded CI env variable; it's already set by the runner
* lowercase `runner.os`


fb55:
Thanks a lot @XhmikosR!

PR #1680, opened 2021-01-16
Filter text nodes in find function

### LINKED ISSUE #1679: rc.4 and rc.5 crash on a script that works in rc.3
To repro:

1. Clone https://github.com/ljharb/es-abstract/ and `npm install`
1. `npm install cheerio@1.0.0-rc.3`
1. `node operations/getOps 2020`, observe it completes
1. `npm install cheerio@1.0.0-rc.4` (or 5)
1. `node operations/getOps 2020`, observe it crashes with:
```sh
$PWD/node_modules/cheerio/lib/api/traversing.js:30
    return newElems.concat(elem.children.filter(isTag));
                                         ^

TypeError: Cannot read property 'filter' of undefined
    at $PWD/node_modules/cheerio/lib/api/traversing.js:30:42
    at Array.reduce (<anonymous>)
    at module.exports.exports.find ($PWD/node_modules/cheerio/lib/api/traversing.js:29:30)
    at Object.<anonymous> ($PWD/operations/getOps.js:29:43)
```

Since this is a dev-only script, I only need to use it in node 14 or higher.

5saviahv:
It seems **lodash** filter returns new empty array even when collection is undefined.

Should fix #1679

### DIFF
--- a/lib/api/traversing.js
+++ b/lib/api/traversing.js
@@ -52,7 +52,9 @@ exports.find = function (selectorOrHaystack) {
   var elems = reSiblingSelector.test(selectorOrHaystack)
     ? context
     : context.reduce(function (newElems, elem) {
-        return newElems.concat(elem.children.filter(isTag));
+        return Array.isArray(elem.children)
+          ? newElems.concat(elem.children.filter(isTag))
+          : newElems;
       }, []);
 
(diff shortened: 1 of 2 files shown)


fb55:
Very nice, thanks!

PR #1681, opened 2021-01-17
add length property for all cheerio instances

### LINKED ISSUE #1674: .merge() - wont accept Cheerio object.
### Test

``` js

const ret = cheerio.merge(new cheerio(), ["elem1","elem2"]);
```
Problem is **merge** uses function **isArrayLike** and **isArrayLike** uses **Object.hasOwnProperty** for testing does object has **length** property. Cheerio objects has **length** property, but it is defined in prototype and it wont meet criteria.

### Option 1

Remove check and let it throw if object has not array like

### Option 2

maybe test is property length a number instead

5saviahv:
* Add length parameter for all cheerio instances.
* re do merge with simple postfix operators

should resolve #1674

### DIFF
--- a/lib/cheerio.js
+++ b/lib/cheerio.js
@@ -40,6 +40,7 @@ var Cheerio = (module.exports = function (selector, context, root, options) {
     return new Cheerio(selector, context, root, options);
   }
 
+  this.length = 0;
   this.options = Object.assign(
     {},
     defaultOptions,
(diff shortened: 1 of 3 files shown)


lib/cheerio.js
```diff
@@ -40,6 +40,7 @@ var Cheerio = (module.exports = function (selector, context, root, options) {
     return new Cheerio(selector, context, root, options);
   }
 
+  this.length = 0;
```
fb55, line 43: Remove the prototype property from L109?

5saviahv: Done

fb55:
This looks great! I should have flagged this earlier — could you include a test (probably based on #1674) to make sure this doesn't break again?

fb55:
Awesome, thanks a lot @5saviahv!

PR #1682, opened 2021-01-17
Export all interfaces in declaration file

### LINKED ISSUE #1540: The Cheerio interface is not exported in the type definitions
See https://github.com/cheeriojs/cheerio/blob/v1.0.0/types/index.d.ts
The interface Cheerio is declared but not exported. It leads to TS compilation errors in projects, that need it. E.G. https://github.com/vuejs/vue-test-utils/issues/1737

### LINKED ISSUE #1649: Feature Request: export Root from the types
Hi,
It would be great if the type Root would be accessible outside. Eg. when I pass around the result of the load method I'd need to use the Root type, the default export is not good in this case.

dominik-korsa:
fixes #1540, fixes #1649 

**Note:** I don't know what I'm doing, but this fix *seems* to be working for my case

In these examples, the code compiles and the console.log output `Hello` as expected
```ts
import cheerio from 'cheerio';

const $: cheerio.Root = cheerio.load('<div>Hello</div>');
console.log($('div').text());
```
```ts
import cheerio, { Root } from 'cheerio';

const $: Root = cheerio.load('<div>Hello</div>');
console.log($('div').text());
```
```ts
import type { Root } from 'cheerio';
import cheerio from 'cheerio';

const $: Root = cheerio.load('<div>Hello</div>');
console.log($('div').text());
```

Tested with this tsconfig: https://gist.github.com/dominik-korsa/5888e24c196e50fa673154e030cf11ce

I wasn't able to test CommonJS (`require()`) syntax

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -274,5 +274,5 @@ declare namespace cheerio {
   }
 }
 
-declare const cheerioModule: cheerio.CheerioAPI;
-export = cheerioModule;
+declare const cheerio: cheerio.CheerioAPI;
+export = cheerio;


5saviahv:
Good catch, it seems it is not good practice to use same variable name twice 🤔

``` typescript
declare namespace cheerio {...}
...
declare const cheerio: cheerio.CheerioAPI;
export = cheerio;
```

rondonjon:
@5saviahv 

> Good catch, it seems it is not good practice to use same variable name twice 

It seems so be pretty common, see [Advanced Combinations](https://www.typescriptlang.org/docs/handbook/declaration-files/deep-dive.html#advanced-combinations) for some examples

rondonjon:
@dominik-korsa, I am happy and also surprised that this works.

rondonjon:
The proposed solution also solves #1692 for me.

Since your solution is more compatible than [#1695](mine), I have closed that one.

keyiis:
@dominik-korsa Are you installed `@types/cheerio`?if use cheerio own types,still not work.

dominik-korsa:
As far as I know `@types/cheerio` is only for versions 0.x, as versions 1.x will include bundled declarations.

keyiis:
@dominik-korsa i used cheerio version
```
"cheerio": "^1.0.0-rc.3"
```

ejose19:
Would be great if this is merged, as exposing interfaces makes it easier to work than ```ReturnType<typeof cheerio.load>``` or similar.

fb55:
Thanks @dominik-korsa!

CzBiX:
This fix doesn't works for me.
Please export all interfaces like this: https://github.com/axios/axios/blob/master/index.d.ts

Also the `import cheerio from 'cheerio'` doesn't work for ES6, see https://stackoverflow.com/a/56348146

jpujol880807:
> image
> 
> This fix doesn't works for me.
> Please export all interfaces like this: https://github.com/axios/axios/blob/master/index.d.ts
> 
> Also the `import cheerio from 'cheerio'` doesn't work for ES6, see https://stackoverflow.com/a/56348146

like @keyiis said downgrade your version of cheerio to `"cheerio": "^1.0.0-rc.3"` and install `@types/cheerio`. Those type definitions work with that version but no with the `1.0.0-rc.5` then u can do `import cheerio from 'cheerio'` and use `cheerio.Root` and so on 
image

Issue #1689, opened 2021-01-20
Cheerio breaks when bundling with Rollup


hansottowirtz:
Due to a circular dependency Rollup emits warnings when bundling Cheerio, and the resulting code is also broken. This wasn't possible in `1.0.0-rc.3` either, but there the problem was in `readable-stream` and `htmlparser2`, which could then be included as external dependencies. In `1.0.0-rc.5` the problem is in Cheerio.

This is caused by `require('./cheerio');` in [lib/static.js#L31](https://github.com/cheeriojs/cheerio/blob/main/lib/static.js#L31)
Somehow this causes `static.js` to be included after `attributes.js`, which relies on `static.js` when doing `var text = require('../static.js').text` in [lib/api/attributes.js#L8](https://github.com/cheeriojs/cheerio/blob/main/lib/api/attributes.js#L8). 

Ideally this would be solved by refactoring Cheerio in ESM, but just removing the circular dependency in `static.js` would also work.

I made a repro here: [hansottowirtz/cheerio-bundling-repro](https://github.com/hansottowirtz/cheerio-bundling-repro).

```
(!) Circular dependencies
node_modules/cheerio/lib/static.js -> node_modules/cheerio/lib/cheerio.js -> node_modules/cheerio/lib/api/attributes.js -> node_modules/cheerio/lib/static.js
node_modules/cheerio/lib/static.js -> node_modules/cheerio/lib/cheerio.js -> node_modules/cheerio/lib/api/attributes.js -> <project>/node_modules/cheerio/lib/static.
js?commonjs-proxy -> node_modules/cheerio/lib/static.js
node_modules/cheerio/lib/static.js -> node_modules/cheerio/lib/cheerio.js -> node_modules/cheerio/lib/api/manipulation.js -> node_modules/cheerio/lib/static.js
```

5saviahv:
Hmm, yes this require in function **.load()** needs fully built Cheerio object. This is why require are inside that function and not top level. It uses basically node module caching ability.

5saviahv:
I managed to use provided cheerio rollup test, but If I test that generated file it fails. So I have no idea, is my modified cheerio build failing or does rollup configuration flawed.

My build does work with all Jest tests so I think configuration, but I am not used rollup before so I am not sure.

hansottowirtz:
> but If I test that generated file it fails

Yes, I made this test to show that the bundling doesn't work. If we could remove the circular dependency the test would not fail.

[...]

PR #1690, opened 2021-01-20
Remove circular dependencies

hansottowirtz:
A draft proposal to remove circular dependencies in Cheerio. It's probably not the prettiest way.

See #1689

### DIFF
--- a/index.js
+++ b/index.js
@@ -8,7 +8,8 @@
  */
 var staticMethods = require('./lib/static');
 
-exports = module.exports = require('./lib/cheerio');
+var Cheerio = require('./lib/cheerio');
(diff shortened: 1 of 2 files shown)


hansottowirtz:
@5saviahv I see you also made a branch, it seems like your proposal makes more sense than this one. Thanks!

5saviahv:
Yes, I made but I need someone to test it, since I dont have experience with rollup.

Also Cheerio webpage is auto generated from JSDoc comments.

fb55:
Thanks a lot @hansottowirtz! I just merged #1691, which should make this redundant.

Issue #1311, opened 2019-05-22
Circular dependencies giving rollup fits


GreenGremlin:
Circular dependencies are valid in cjs, but create all kinds of issues when converted to es6. I'm not familiar with the Cheerio code, but wanted to open this issue to see how feasible it may be to factor out Cheerio's circular dependencies.

### Rollup output (slightly re-formatted)

```
Circular dependency: node_modules/cheerio/lib/cheerio.js ->
        node_modules/cheerio/lib/api/attributes.js ->
        node_modules/cheerio/lib/static.js ->
        node_modules/cheerio/lib/cheerio.js
Circular dependency: node_modules/cheerio/lib/cheerio.js ->
        node_modules/cheerio/lib/api/attributes.js ->
        node_modules/cheerio/lib/static.js ->
        commonjs-proxy-/Users/jonathan/my-package/node_modules/cheerio/lib/cheerio.js ->
        node_modules/cheerio/lib/cheerio.js
```

fb55:
Happy to accept PRs :)

Sureiya:
This doesn't exist in 1.0.0 branch but does exist in 1.0.0rc.3,  there is another circular dependency happening with html2parser versions less than 4.0.0 as well.

tobiasmuehl:
Any updates on this?

fb55:
Just published a new release, which should resolve this if it is resolved on the master branch.

tvanc:
I'm still getting this exact issue on 1.0.0-rc5

5saviahv:
It is not fixed in 1.0.0-rc5 yet

fb55:
Fixed in RC6.

PR #1691, opened 2021-01-20
remove node module caching dependency

5saviahv:
- move **.load()** function own file so api methods could access other static functions early.
- load **load.js** file after cheerio object is built.

should solve #1311 & #1689

it passes rollup test


fb55 (approved):
One outdated comment, otherwise this looks good!

PR #1701, opened 2021-01-27
add property "textContent"

### LINKED ISSUE #1206: el.prop("tagName") throws error when el is a text node?
Text nodes don't seem to have a "name" property, so this line throws an error since there is check for that (at least it did for me)

https://github.com/cheeriojs/cheerio/blob/master/lib/api/attributes.js#L126

```
     case 'tagName':
     case 'nodeName':
     property = this[0].name.toUpperCase();
```

5saviahv:
- since `text()` function wont return **style** & **script** tag content, so this will give option by mimicking [textContent](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent) property.
- PR should also resolve #1206

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -6,6 +6,7 @@
  */
 
 var text = require('../static').text;
+var html = require('../static').html;
 var utils = require('../utils');
 var isTag = utils.isTag;
 var domEach = utils.domEach;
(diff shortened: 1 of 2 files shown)


5saviahv:
Yes, I used **for..of** loops, but I tested it in node 4 and according to [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of) it should work even with 0.12

PR #1704, opened 2021-01-29
Allow foreach loops in typescript

mcpiroman:
Reflects missing ts declaration of this line:
https://github.com/cheeriojs/cheerio/blob/897b37fb671382e8f804d8f1f15e379964f48ad2/lib/cheerio.js#L139


fb55:
Good catch, thanks!

PR #1706, opened 2021-02-01
root will be overwritten

5saviahv:
root will be overwritten if you parse new html content.

lets say you 
``` js
const $ = cheerio.load(html);

const el = $('<div></div>').add('#fruits');
// should return 2 elements but instead you get just one
```
turns out since `el._root` & `el` are same instance. if you add `div` into dom you effectively overwriting `root`. Now when you plan make query against that old root you get empty results since new empty `div` wont have old content.

### DIFF
--- a/lib/cheerio.js
+++ b/lib/cheerio.js
@@ -77,8 +77,8 @@ var Cheerio = (module.exports = function (selector, context, root, options) {
   }
 
   // If we don't have a context, maybe we have a root, from loading
-  if (!context && this._root && this._root.length) {
-    context = Cheerio.merge(this, this._root);
+  if (!context) {
+    context = this._root;
   } else if (typeof context === 'string') {
     if (isHtml(context)) {
       // $('li', '<ul>...</ul>')
--- a/lib/cheerio.js
+++ b/lib/cheerio.js
@@ -53,7 +53,7 @@ var Cheerio = (module.exports = function (selector, context, root, options) {
 
   if (root) {
     if (typeof root === 'string') root = parse(root, this.options, false);
-    this._root = Cheerio.call(this, root);
+    this._root = new Cheerio(root);
   }
 
   // $(<html>)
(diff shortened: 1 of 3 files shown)


lib/cheerio.js
```diff
@@ -77,8 +77,8 @@ var Cheerio = (module.exports = function (selector, context, root, options) {
   }
 
   // If we don't have a context, maybe we have a root, from loading
-  if (!context) {
-    context = this._root;
+  if (!context && this._root && this._root.length) {
+    context = Cheerio.merge(this, this._root);
```
fb55, line 81: Not sure I fully understand this change.

fb55:
@5saviahv This is yet another great PR. I have one question about this, otherwise it should have been pretty much good to go.

lib/cheerio.js
5saviahv: Main coal was prevent object `_root` be returned. if you save it directly to context it will returned later and `this` and `_root` will be actually be same object. This way only data is transferred.

5saviahv: if context was not provided, but root was provided - `this` becomes context + we add data from root. So even when `_root` are still Cheerio object like before they are not same object.

5saviahv:
My doubts were mainly if context is provided and is object - do I have to add some more testing there ?

lib/cheerio.js
fb55: `this` should always be empty (see the `return` in L76), so I don't think this is necessary.

5saviahv: I am not sure I understand: `this` on L76 is no longer empty. but starting from L80 we start  building `context` variable. In the end  `context` has to be cheerio object or we cant execute `.find` and in the process empty `this` is very much discarded (L101).

fb55: On L80, `this` shouldn't have any values, as we immediately return after the only time where elements are added to it (on L76). That means this is equivalent:

```suggestion
  if (!context) {
    context = this._root;
```

5saviahv: hmm, actually it is not equivalent, `this` actually becomes clone of `this._root`, but so far no one complains so I relent.

lib/cheerio.js
```diff
@@ -53,7 +53,7 @@ var Cheerio = (module.exports = function (selector, context, root, options) {
 
   if (root) {
     if (typeof root === 'string') root = parse(root, this.options, false);
-    this._root = Cheerio.call(this, root);
+    this._root = new Cheerio(root);
```
fb55, line 56: Finally figured out what's wrong here: This should be `new this.constructor(root)`. Incorporating this in a different PR.

Issue #1105, opened 2017-11-05
Can't reach anything under noscript tag


rightaway:
I've got `cheerio.load(pageContent, { decodeEntities: false })`, but then when I try to match an element that's under a `noscript` tag it doesn't work. I thought setting `decodeEntities` to `false` would allow this. How do I match that element?

VulcanRav:
try cheerio.load(pageContent, [ xmlMode: true});

monkeysuffrage:
I'm seeing this too. This should not have been closed.

fb55:
Enabling xml mode fixes this, otherwise parse5 will always strip `noscript` tags.

monkeysuffrage:
Right, but it's still an issue, right? It should stay open until it gets resolved.

mitin001:
Try to match the noscript tag itself, get the html by calling .html() and then load that html with cheerio again. That way you'll be able to match any element under the noscript tag.

paulsmithkc:
This is definitely still a bug with cheerio, as cheerio is essentially a browser without JavaScript support. The noscript tag is intended to provide content for browsers that do not support javascript (which would include search engines, web crawlers, and web scrapers).

"xmlMode: true" has several other side effects (see documentation) which can cause most pages, and especially those in question to fail to parse.

joeybaker:
Quickly looking through, it seems like https://github.com/cheeriojs/cheerio/blob/208bce1ee8ed921dbd0fc2988644fd3a68bf8bd1/lib/parse.js needs to be updated to turn `scriptingEnabled: false` https://github.com/inikulin/parse5/blob/master/packages/parse5/docs/options/parser-options.md 

I'm not sure if there are security implications of doing that though.

paulsmithkc:
<noscript> tags should be parsed as HTML as that is what they contain, not text/javascript.

dsmmcken:
This is the first google result for "cheerio noscript". If you are attempting to parse html inside a noscript tag with cheerio, see the doc page linked below.

Set the [`scriptingEnabled`](https://cheerio.js.org/interfaces/Parse5Options.html) option to `false`.

```
const $ = cheerio.load(data, { scriptingEnabled: false });
```

ggorlen:
[`xmlMode`](https://cheerio.js.org/interfaces/CheerioOptions.html#xmlMode) is deprecated in favor of [`xml`](https://cheerio.js.org/interfaces/CheerioOptions.html#xml).

[...]

PR #1707, opened 2021-02-01
expose parse5 option scriptingEnabled

5saviahv:
Expose [scriptingEnabled](https://github.com/inikulin/parse5/blob/master/packages/parse5/docs/options/parser-options.md#optional-scriptingenabled) option. It would accept `false` value only, all other cases it should stay true.

``` html
<noscript><link rel="stylesheet" href="/css/noscript.css"></noscript>
```
It should help parse some documents where `noscript` elements are used. 
If **true** (_default_) all html code inside `noscript` elements will treated as text,
if **false** all html code inside `noscript` elements will treated as html and parsed accordingly.

#1105

### DIFF
--- a/lib/parsers/parse5.js
+++ b/lib/parsers/parse5.js
@@ -4,6 +4,10 @@ var htmlparser2Adapter = require('parse5-htmlparser2-tree-adapter');
 
 exports.parse = function (content, options, isDocument) {
   var opts = {
+    scriptingEnabled:
+      typeof options.scriptingEnabled === 'boolean'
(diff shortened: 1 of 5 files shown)


fb55:
This looks great! Could you (1) add a simple test case and, optionally, (2) add this to the TS type def?

5saviahv:
My doubts were mainly if context is provided and is object - do I have to add some more testing there ?

fb55:
Thanks a lot!

PR #1708, opened 2021-02-01
small parentsUntil fixes

5saviahv:
- parentsUntil selector accepts multiple results.
- parentsUntil filter didnt filter

### DIFF
--- a/lib/api/traversing.js
+++ b/lib/api/traversing.js
@@ -152,11 +152,11 @@ exports.parentsUntil = function (selector, filter) {
   var untilNodes;
 
   if (typeof selector === 'string') {
-    untilNode = select.select(
+    untilNodes = select.select(
       selector,
       this.parents().toArray(),
       this.options
-    )[0];
+    );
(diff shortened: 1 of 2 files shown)


fb55:
@5saviahv This is a great find, any reason you closed the PR?

5saviahv:
I was panicking when one of my tests fail. Turns out I had typo in tests and it wasn't even pushed github.

lib/api/traversing.js
```diff
@@ -152,11 +152,11 @@ exports.parentsUntil = function (selector, filter) {
   var untilNodes;
 
   if (typeof selector === 'string') {
-    untilNode = select.select(
+    untilNodes = select.select(
```
fb55, line 155: Shouldn't this be a `filter` as well?

5saviahv: now when you say it ... yes it should.

fb55:
Nice stuff!

PR #1711, opened 2021-02-01
Fix wrong method used in `.map` tests

### LINKED ISSUE #1705: Test for .map method is actually calling .each method
Hello,

I was looking at some test files and I found this line that I think is wrong:

https://github.com/cheeriojs/cheerio/blob/9b28b49527b22f72ef81e28c069302ca3eb7c329/test/api/traversing.js#L706

It's in the tests for the `.map` method but has been changed to `.each` in https://github.com/cheeriojs/cheerio/commit/21de2c55d0459db0564de91fed20fd33c9647e19

Not sure why it's been changed, but it should be `.map` being tested here right?

I can open a PR if requested.

Pustur:
The added `return` is to prevent linter errors due to the rule `array-callback-return` (which is probably why the `.map` call was originally replaced with `.each`).

Closes #1705.

### DIFF
--- a/test/api/traversing.js
+++ b/test/api/traversing.js
@@ -703,9 +703,10 @@ describe('$(...)', function () {
       var args = [];
       var thisVals = [];
 
-      $fruits.each(function () {
+      $fruits.map(function () {
         args.push(Array.prototype.slice.call(arguments));
         thisVals.push(this);
+        return;
       });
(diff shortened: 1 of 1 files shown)


fb55:
Great find @Pustur, and thanks a lot for fixing this!

PR #1716, opened 2021-02-04
refactor(utils): Simplify `quickExpr`

fb55:
Now allows a `#` before the string.

BREAKING: Now requires a letter directly after the `<`. Both parsers would have parsed this as text before.

@5saviahv #1699 was a great find, thanks for opening it! I wasn't sure about the actual implementation, here is my take; with it, we still allow text around the actual string, but ban things that definitely aren't tags (see breaking).

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -84,8 +84,13 @@ exports.cloneDom = function (dom) {
   return clone;
 };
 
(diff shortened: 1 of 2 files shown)


5saviahv:
I see whats your aim is, and since it acts only for testing, it should be fine.

PR #1720, opened 2021-02-05
Prettier: JSDoc comments needs formatting

5saviahv:
new Prettier JSDoc plugin version reformats long comments and linter do fail now because it.


5saviahv:
sry prettier plugin different versions seems to conflicting with each other

PR #1721, opened 2021-02-08
add after, before, slice arguments into jsdoc.

5saviahv:
according to documentation functions [after](https://cheerio.js.org/Cheerio.html#after), [before](https://cheerio.js.org/Cheerio.html#before), [slice](https://cheerio.js.org/Cheerio.html#slice) doesnt have arguments, added argument definitions.

### DIFF
--- a/lib/api/manipulation.js
+++ b/lib/api/manipulation.js
@@ -602,7 +602,7 @@ exports.insertAfter = function (target) {
  *   //      <li class="pear">Pear</li>
  *   //    </ul>
  *
- * @param {string | Element | Element[] | Cheerio | Function} content - HTML
+ * @param {...(string | Element | Element[] | Cheerio | Function)} content - HTML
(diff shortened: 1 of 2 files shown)


lib/api/manipulation.js
```diff
@@ -593,6 +602,9 @@ exports.insertAfter = function (target) {
  *   //      <li class="pear">Pear</li>
  *   //    </ul>
  *
+ * @param {string | Element | Element[] | Cheerio | Function} content - HTML
```
fb55, line 605: Not sure how to do this properly, but we should use a [repeatable param](https://jsdoc.app/tags-param.html#multiple-types-and-repeatable-parameters) here.

```suggestion
 * @param {...(string | Element | Element[] | Cheerio | Function)} content - HTML
```

fb55:
Nice PR!

PR #1722, opened 2021-02-08
Object.assign is not compatible with node v0.12

5saviahv:
After writing out `lodash` - function `_.assign` was replaced with `Object.assign`, but this is not compatible with node v0.12

### DIFF
--- a/lib/api/attributes.js
+++ b/lib/api/attributes.js
@@ -246,7 +246,7 @@ function setData(el, name, value) {
     el.data = {};
   }
 
-  if (typeof name === 'object') Object.assign(el.data, name);
+  if (typeof name === 'object') utils.assign(el.data, name);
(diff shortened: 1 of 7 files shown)


5saviahv:
Probably not good idea since it reduces code coverage.

TrySound:
@5saviahv Curious, why do you want to support node 0.12?

5saviahv:
I have no preference, but Cheerios minimum supported version is currently chosen to be 0.12

Simply using `Object.assign` Cheerios minimum version should be raised to 4.0.0 - I have tested cheerio with 4.0.0 and it works but testing with 0.12 fails.

fb55:
I am also starting to be more and more in favor of bumping the minimum version to Node 4. I do think there is value in supporting old Node versions, but there are only so many hoops worth jumping through for it.

PR #1728, opened 2021-02-11
fix(traversing): Fix filter for `{prev,next}Until`

fb55:
Continuation of #1708, fixing up `untilNodes` for `{prev,next}Until`.

### DIFF
--- a/lib/api/traversing.js
+++ b/lib/api/traversing.js
@@ -181,9 +181,9 @@ exports.parentsUntil = function (selector, filter) {
       }
     }, this);
 
-  return this._make(
-    filter ? select.filter(filter, parentNodes, this.options) : parentNodes
-  );
+  return filter
+    ? exports.filter.call(parentNodes, filter, this)
+    : this._make(parentNodes);
 };
 
 /**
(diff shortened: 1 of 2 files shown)


lib/api/traversing.js
```diff
@@ -181,9 +181,9 @@ exports.parentsUntil = function (selector, filter) {
       }
     }, this);
 
-  return this._make(
```
fb55, line 184: Updated this to match the pattern of most other functions.

fb55:
@5saviahv I know that #1662 is still open (which would fix all similar issues), but this seems worth its own fix.

lib/api/traversing.js
5saviahv: hmm, yes you are correct other traversing functions use filter function like that.

lib/api/traversing.js
```diff
@@ -314,7 +314,7 @@ exports.nextUntil = function (selector, filterSelector) {
   var untilNodes;
 
   if (typeof selector === 'string') {
-    untilNode = select.select(selector, this.nextAll().get(), this.options)[0];
+    untilNodes = this.nextAll(selector).toArray();
```
5saviahv, line 317: nice

lib/api/traversing.js
```diff
@@ -432,7 +432,7 @@ exports.prevUntil = function (selector, filterSelector) {
   var untilNodes;
 
   if (typeof selector === 'string') {
-    untilNode = select.select(selector, this.prevAll().get(), this.options)[0];
+    untilNodes = this.prevAll(selector).toArray();
```
5saviahv, line 435: nice

5saviahv (approved):
Nice, I thought those functions should also be updated.

PR #1730, opened 2021-02-11
cheerio-select 1.1.0 -> 1.2.0

5saviahv:
Same as #1729 except I updated failed test

apparently order of objects should be correct this time.


fb55:
Awesome, thanks! I was a bit confused by the failing test, thanks for clearing things up.

PR #1734, opened 2021-02-13
nesting level of some deeply nested tests decreased

5saviahv:
`.add` function had some 3rd and 4th level nested tests.
- looking from output those indentations are rather subtle and no longer easily readable, I moved them to 2nd level and added argument descriptions used in tests.
- Also those deeply nested tests aren't very easy to follow in code.
- no actual change in tests, just formatting.

### DIFF
--- a/test/api/traversing.js
+++ b/test/api/traversing.js
@@ -1155,7 +1155,7 @@ describe('$(...)', function () {
     });
   });
 
-  describe('.add', function () {
+  describe('.add()', function () {
     var $fruits;
     var $apple;
     var $orange;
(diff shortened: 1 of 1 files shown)


test/api/traversing.js
```diff
@@ -1346,79 +1342,80 @@ describe('$(...)', function () {
...
-        it('occur within the current selection', function () {
-          var $selection = $fruits.add($('#fruits li'));
-
-          expect($selection).toHaveLength(4);
-          expect($selection[0]).toBe($fruits[0]);
-          expect($selection[1]).toBe($apple[0]);
-          expect($selection[2]).toBe($orange[0]);
-          expect($selection[3]).toBe($pear[0]);
-        });
+    //d escribe('(selection) :', function () {
```
fb55, line 1345:
```suggestion
```

fb55:
Nice PR! One comment seems to have been forgotten, otherwise this should be good to go.

fb55:
Thanks for following up!

PR #1740, opened 2021-02-16
add wrapAll into types

5saviahv:


fb55:
I actually had your PR from a while ago bookmarked, that already added this 😄

Thanks for following up!

PR #1747, opened 2021-02-19
mix text and html tags in tests

5saviahv:
- updated some tests, so they will try functions with mixed text and html tags
- updated addclass function test so it will try elements without already existing class name


fb55:
Very nice, thanks a lot @5saviahv!

PR #1757, opened 2021-02-25
function .attr() - handle value undefined like jQuery 

### LINKED ISSUE #1664: `attr` converts `undefined` to string
jQuery removes attributes when setting their value to `undefined`, while cheerio sets them to the string `'undefined'`.

Discovered in #1659.

5saviahv:
function .attr() - handle value undefined like jQuery does it.

should fix #1664


fb55:
Nice patch, thanks!

Issue #1591, opened 2020-12-24
content.ForEach is not a function


mrjackyliang:
Hello, I am a developer of a Homebridge plugin, and I am having a user report this error:

```
[12/23/2020, 4:09:41 PM] TypeError: content.forEach is not a function
    at parse (/homebridge/node_modules/homebridge-adt-pulse/node_modules/cheerio/lib/parse.js:38:15)
    at Function.exports.load (/homebridge/node_modules/homebridge-adt-pulse/node_modules/cheerio/lib/static.js:39:14)
    at /homebridge/node_modules/homebridge-adt-pulse/api.js:598:39
    at arrayMap (/homebridge/node_modules/homebridge-adt-pulse/node_modules/lodash/lodash.js:639:23)
    at Function.map (/homebridge/node_modules/homebridge-adt-pulse/node_modules/lodash/lodash.js:9580:14)
    at Request._callback (/homebridge/node_modules/homebridge-adt-pulse/api.js:597:28)
    at Request.self.callback (/homebridge/node_modules/homebridge-adt-pulse/node_modules/request/request.js:185:22)
    at Request.emit (events.js:315:20)
    at Request.<anonymous> (/homebridge/node_modules/homebridge-adt-pulse/node_modules/request/request.js:1154:10)
    at Request.emit (events.js:315:20)
[12/23/2020, 4:09:41 PM] Got SIGTERM, shutting down Homebridge...
```

More details in [this issue](https://github.com/mrjackyliang/homebridge-adt-pulse/issues/39)

5saviahv:
Cheerio.load expects there to be array. Older versions was there condition for checking if it not array. Seems like it is lost due to optimizations.

mrjackyliang:
Yeah, the input is an array. I don't get why it's having this error.

5saviahv:
sample from your code where it happens
``` javascript
...
const $ = cheerio.load(body);
const sensors = $('#orbSensorsList table tr.p_listRow').toArray();

const output = _.map(sensors, (sensor) => {
  const theSensor = cheerio.load(sensor);
  const theName = theSensor('a.p_deviceNameText').html();
  ...
})
...
```
variable `sensor` is single object (html node)

older version wraps it into array

mrjackyliang:
Oh haha! Totally missed that actually.

mrjackyliang:
[...]

PR #1759, opened 2021-02-25
Fix types

karlhorky:
Hi there, first of all, thanks so much for Cheerio! Cool to be able to load an HTML string in anywhere and traverse so easily.

Shouldn't `.get()` always return either `Element[]`, an `Element` or `undefined?

### DIFF
--- a/types/index.test-d.ts
+++ b/types/index.test-d.ts
@@ -1,4 +1,5 @@
 import { expectType } from 'tsd';
+import { Element } from 'domhandler';
 import cheerio from '..';
 
 /*
@@ -19,7 +20,7 @@ cheerio(html);
 cheerio('ul', html);
 cheerio('li', 'ul', html);
 
-const $fromElement = cheerio.load($('ul').toString());
+const $fromElement = cheerio.load($('ul').get(0) as Element);
 
 if ($fromElement('ul > li').length !== 3) {
   throw new Error(
--- a/types/index.test-d.ts
+++ b/types/index.test-d.ts
@@ -19,7 +19,7 @@ cheerio(html);
 cheerio('ul', html);
 cheerio('li', 'ul', html);
 
-const $fromElement = cheerio.load($('ul').get(0));
+const $fromElement = cheerio.load($('ul').toString());
 
 if ($fromElement('ul > li').length !== 3) {
   throw new Error(
@@ -248,7 +248,7 @@ $el.eq(0).text();
 $el.eq(-1).text();
 
 // .get( [i] )
-$el.get(0).tagName;
+$el.get(0)?.tagName;
 $el.get().length;
 
 // .index()
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -125,8 +125,8 @@ declare namespace cheerio {
 
     eq(index: number): Cheerio;
 
-    get(): any[];
-    get(index: number): any;
+    get(): Node[];
+    get(index: number): Node | undefined;
 
     index(): number;
     index(selector: string): number;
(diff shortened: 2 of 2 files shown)


types/index.test-d.ts
```diff
@@ -19,7 +19,7 @@ cheerio(html);
 cheerio('ul', html);
 cheerio('li', 'ul', html);
 
-const $fromElement = cheerio.load($('ul').get(0));
+const $fromElement = cheerio.load($('ul').toString());
```
5saviahv, line 22: I believe idea here was test does "cheerio.load" accepts type "Element" as input, but I also did see error with last commit, so maybe cast it instead:
``` js
const $fromElement = cheerio.load($('ul').get(0) as Element);
```

I know type "Element" is not defined, but we could import it

``` js
import type { Element } from 'domhandler';
```

5saviahv (approved):
Nice work

types/index.test-d.ts
karlhorky: That code is actually broken, even in JavaScript:

```js
const $fromElement = cheerio.load($('ul').get(0));
```

It throws an error. `cheerio.load()` doesn't accept an `Element`.

5saviahv: Interesting, I tested it and tsd didn't complain. I also tested load function with  and it worked, like expected.

``` javascript
const html = 
  '<ul id="fruits">\n<li class="orange">Apple</li>\n<li class="class">Orange</li>\n<li class="pear">Pear</li>\n<input type="text" />\n</ul>';

const $ = cheerio.load(html);
const $fromElement = cheerio.load($('li').get(0));

console.log(cheerio.html($fromElement._root));
```
I copied what you did and added casting [cb6f40](https://github.com/cheeriojs/cheerio/commit/cb6f40f8607161c48405e975b03b79085e99aec6) and it works. Interesting why it fails on you.

karlhorky: I'm not talking about tsd / TypeScript though - did you try actually running that as JavaScript? `cheerio.load()` with an element throws an error. The type for the `load` method also seems to indicate that it does not accept an element.

karlhorky: Ah sorry I think I was looking at the wrong overload. It does seem to accept an element here: https://github.com/cheeriojs/cheerio/blob/1c2c0c0a86f60da50c03bbcd192b8e671178a3b6/types/index.d.ts#L282

karlhorky: I've added the cast now, if you would prefer that over the `.html()` method.

5saviahv: I have to admit I rarely use typescript, I do see why it is useful but still. I know there was issue with element (still are, rc.5 in npm) #1591, but it should be fixed in repo.

karlhorky: Right, yeah, that must have been the error that was thrown.

karlhorky: And as for the value of TypeScript, it's actually telling us a useful bit of information here:

If the element returned from `.get(0)` is `undefined`, then there will be an error with `cheerio.load()`.

So actually, a more robust version of the code would be this:

```ts

const $ul = $('ul').get(0);

if ($ul) {
  const $fromElement = cheerio.load($ul);

  if ($fromElement('ul > li').length !== 3) {
    throw new Error(
      'Expecting 3 elements when passing `cheerio.Element` to `load()`'
    );
  }
}

```

But since these are just tests, it doesn't matter too much that there is a cast here.

fb55:
Thanks a lot @karlhorky!

karlhorky:
Glad to help :) Thanks for the merge!

PR #1788, opened 2021-03-28
Add type for $.text() method

karlhorky:
It appears there is a `$.text()` method that is not currently typed (at least this form: `text(): string;`).

Not sure if the `text(text: string): Cheerio;` form works... copied that from the Cheerio element declaration above.

### DIFF
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -262,7 +262,7 @@ declare namespace cheerio {
       context?: Document,
       keepScripts?: boolean
     ): Document[];
-    
+
     text(): string;
     text(text: string): Cheerio;
 
--- a/types/index.d.ts
+++ b/types/index.d.ts
@@ -262,6 +262,9 @@ declare namespace cheerio {
       context?: Document,
       keepScripts?: boolean
     ): Document[];
+    
+    text(): string;
+    text(text: string): Cheerio;
 
     html(options?: CheerioParserOptions): string;
     html(


fb55:
The entire module has been converted to TS, so this is hopefully resolved! Thanks @karlhorky.

karlhorky:
Oh nice, do you know which file that would be in?

When will the TS version be released? I can then remove this patch from my codebase (currently I'm applying it using [`patch-package`](https://www.npmjs.com/package/patch-package))

fb55:
You can look at the new docs, which includes a link to the source code! https://cheerio.js.org/classes/cheerio.html#text

The source link actually links to an alias, here is the actual function definition: https://github.com/cheeriojs/cheerio/blob/463d4049d7fb382a8160011b95e0ea1c6c9a94f2/src/static.ts#L136

karlhorky:
Great, these docs are a huge step up!

Ah I see there's a `1.0.0-rc.7` version - are the new types already in there? I'll try it out :)

fb55:
They are, I'm currently writing the release notes :)

karlhorky:
Ah, seems like `lib` folder is missing from the published package? https://unpkg.com/browse/cheerio@1.0.0-rc.7/

fb55:
Oooff, thanks for flagging! Pushed out `1.0.0-rc.8` that fixes this.

karlhorky:
No worries, just tried out `1.0.0-rc.8`, and still some problems - slightly different but similar:

1. `package.json` specifies `"main": "lib/index.js", "types": "lib/index.d.ts"`, both of which don't exist: https://unpkg.com/browse/cheerio@1.0.0-rc.8/lib/
2. there's an extra `package.json` in `lib`: https://unpkg.com/browse/cheerio@1.0.0-rc.8/lib/package.json

fb55:
Yeah, just noticed. For some reason, `tsc` is considering the `package.json` as a source as well, and adds it to the resulting `lib` directory.

fb55:
Okay, turns out we were exporting the `version` field from the `package.json`, which made TypeScript include the `package.json` in the output. I've removed the field, and the build now looks as expected.

karlhorky:
Ok this is looking good now, thanks!

Now I just have to fix all the other type errors that have appeared in my project :D

karlhorky:
Ok, maybe this one type error would be worth it for me to open a new PR to Cheerio actually: https://github.com/cheeriojs/cheerio/issues/1844

PR #1797, opened 2021-03-31
Parsing HTML content inside <noscript> tags

paulsmithkc:
The `scriptingEnabled` flag was added to parse5 in version 5.0.0 [parse5: ParserOptions](https://github.com/inikulin/parse5/blob/master/packages/parse5/docs/options/parser-options.md)  

`scriptingEnabled=true` will parse `<script>` tags as javascript and `<noscript>` tags as raw text.
`scriptingEnabled=false` will parse `<script>` tags as raw text and `<noscript>` tags as HTML.

The later is the preferred default behavior for cheerio. As we do not want to execute the javascript, but do want to view the page as a scripts-disabled browser would.

See #1105 for discussion on this issue.

### DIFF
--- a/lib/load.js
+++ b/lib/load.js
@@ -28,6 +28,7 @@ exports.load = function (content, options, isDocument) {
   }
 
   options = Object.assign({}, defaultOptions, flattenOptions(options));
+  options.scriptingEnabled = !!options.scriptingEnabled;
 
   if (typeof isDocument === 'undefined') isDocument = true;
 
--- a/lib/options.js
+++ b/lib/options.js
@@ -3,6 +3,7 @@
 exports.default = {
   xml: false,
   decodeEntities: true,
+  scriptingEnabled: false,
 };
 
 var xmlModeDefault = { _useHtmlParser2: true, xmlMode: true };
--- a/test/cheerio.js
+++ b/test/cheerio.js
@@ -428,35 +428,38 @@ describe('cheerio', function () {
   describe('parse5 options', function () {
     var noscript = fixtures.noscript;
 
-    // should parse noscript tags only with false option value
+    // should parse noscript tags by default
     test('{scriptingEnabled: ???}', function () {
       var opt = 'scriptingEnabled';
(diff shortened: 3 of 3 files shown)


5saviahv:
I can't understand why change Cheerios current default behaviour `scriptingEnabled=true` to `scriptingEnabled=false`.

paulsmithkc:
The default behavior is not the correct or expected behavior. It doesn't make sense for users to need to opt-out of bad/abnormal behavior.

5saviahv:
I am not against it but I simply dont like to change current default behaviour, since you have already option to explicitly pass in option `scriptingEnabled=false`.

paulsmithkc:
But that behavior also is not documented anywhere in cheerio's docs. I checked. @joeybaker found it buried in a parse5.

5saviahv:
true, documentation should be better.

fb55:
I don't think making this the default behavior is the right choice here. The current documentation should make it easier to find the option. Happy to accept further updates to the docs to make it clear.

joeybaker:
@fb55 why do you believe this is the wrong default behavior? I'd think it's unlikely that cheerio users will not want `<noscript>` tags parsed as HTML?

paulsmithkc:
@fb55 Both me and @joeybaker agree that we expect HTML content of <noscript> to be parsed as HTML and that the current behavior is **unexpected.**

Perhaps it would be better to ask what your users/customers expect?

Personally, I **expect** cheerio to behave like a JS disabled browser. I expect it to parse **all** of the HTML. Not to pick and choose what HTML it parses.

5saviahv:
Cheerio is not a web browser ... it simply tries to give a similar results than jQuery in browser. 

I dont think majority will expect Cheerio bring some additional tags into dom if they load web page with Cheerio. You expect same results than jQuery gives you.

In concept Cheerio acting like "JS disabled browser" maybe good, but making it default behaviour will be **breaking change** with no good reason.

Issue #1585, opened 2020-12-22
rc.4 includes an implicit Node bump


ljharb:
When I updated to rc5, tests started failing in node 4: https://travis-ci.com/github/enzymejs/enzyme/builds/210326410

Specifically, parse5-htmlparser2-tree-adapter is using object destructuring syntax, which means including it is a breaking change :-/

fb55:
Thanks for flagging! For starters, it is unfortunate that this wasn't caught earlier. The CI should cover the oldest supported Node version, which it didn't do here. Sorry about that!

It seems like parse5 doesn't support node@4 anymore. This also applies to eg. [the serializer](https://github.com/inikulin/parse5/blob/9c683e154b8bbdc94237580680a934348406d1a2/packages/parse5/lib/serializer/index.js), which uses `class` syntax. Same for files used by the parser.

I'm inclined to bump the minimal version requirement of cheerio, as not using parse5 is not really an option at this point. Given that we are about to release a major version, this would be the point in time to do it. And when we do the bump, we might just go to the current Node LTS version, which is 10.

@ljharb I'd love to hear your thoughts here. I know this is an inconvenience for enzyme, and I'd love find a good solution for both projects.

TrySound:
@fb55 rc releases should've probably be published with --tag=next
@ljharb It's very sad we had to have bloated node_modules because of many years dead node versions in all your projects

fb55:
> rc releases should've probably be published with --tag=next

Agreed. Unfortunately, changing this now will leave people stuck on one of the older RCs, as well as promote an old version that features many issues solved in more recent releases. Also see #1261.

> It's very sad we have to have bloated node_modules because of many years dead node versions in all your projects

That's a bit harsh. There is a lot of value in supporting as wide of a user base as possible, and the success these projects have speaks for itself. That's why I also don't want to be too aggressive with cheerio.

I'd love to maintain backwards compatibility to Node 4, but don't see a great way forward given that we depend on parse5. 

One more involved way would be to separate parsers/serializers out from cheerio. @matthewmueller proposed something similar in https://github.com/cheeriojs/cheerio/discussions/1271#discussioncomment-207120. We could move parsers & serializers to the options, allowing users with older systems (or size limits) to use `htmlparser2`.

ljharb:
@TrySound Testing frameworks in particular need to be able to test in all supported environments, which includes browsers for which older node versions are a good substitute. node_modules size is unimportant; supporting users is quite important.

[...]

PR #1816, opened 2021-04-16
refactor: Convert to TypeScript

fb55:
Initial big lift to convert the codebase, broken into three commits to make the bigger changes obvious. Everything seems to work now, including detection if users pass selectors (`$('<div>')` will product a `Cheerio<Node>` type, `$('.foo')` a `Cheerio<Element>`).

Currently missing:
- [x] The ergonomics of using types aren't great. We should re-export all regularly used types.
- We can now use `typedoc` to generate our documentation. Produced docs still have to be improved:
  - [x] Examples now need to be put in code blocks.
  - [x] The organization is very hard to grasp.
- [x] Benchmarks have to be updated to use `ts-node`.


tsconfig.json
```diff
@@ -1,5 +1,37 @@
 {
   "compilerOptions": {
-    "esModuleInterop": true
-  }
+    /* Basic Options */
+    "target": "es5" /* Specify ECMAScript target version: 'ES3' (default), 'ES5', 'ES2015', 'ES2016', 'ES2017', 'ES2018', 'ES2019' or 'ESNEXT'. */,
```
XhmikosR, line 4: @fb55 sorry for the CC on an old PR, but I didn't want to make a new issue just for this. Is there any reason you didn't go with `ES2017` or whatever is equivalent for Node.js 10, which is the oldest supported Node.js version?

XhmikosR: Ah, NVM I see now that the package.json engines is set to `>=6`. On CI Node.js >= 10 is tested hence why I thought this could be improved. IMHO, these two should be in sync, especially since Node.js 6 is ancient...

fb55: Jest doesn't support Node 6, which is why Node 6 isn't part of the CI anymore.

I am still a bit hesitant to bump the Node version, primarily due to the discussion in #1585. But the mismatch between CI and `package.json` seems like a good reason to make the cut.

XhmikosR: Ah, now I remember, thanks for the info.

So, yeah, not sure what the proper semver way would be. Perhaps release v1.0.0 final with the current engines, and then cut a new v2.0.0 with the bump in engines and tsconfig.

PR #1830, opened 2021-04-30
Fixed parsing content in .html function

aletorrado:
Hi, I've noticed that calling .html(<content>) when using the `htmlparser2` mode will crash. This will fix that. Thanks


fb55:
Closing this as it has been fixed in #1816. Thanks a lot @aletorrado!

Issue #1844, opened 2021-05-06
Accept Uint8Array also instead of Buffer?


karlhorky:
Hi @fb55 ! In `1.0.0-rc.9` I noticed that [the types for `cheerio.load`](https://github.com/cheeriojs/cheerio/blob/726f94a4677fd07b9528079172fe6176e70235be/src/load.ts#L25) only accept `string | Node | Node[] | Buffer`.

I'm trying to pass in a `vfile.VFileContents` (via `remark.process`), which is `string | Uint8Array`.

Would it be possible to also accept [`Uint8Array`, the base type for `Buffer`](https://github.com/vfile/vfile/blob/f95280cbd0d8d29a3fd49984e91e7c665175bceb/lib/index.js#L16)?

fb55:
Happy to accept PRs for this! We currently rely on `Buffer.isBuffer` to see if something is a buffer; if you have a more generic approach, I'm happy to take it :)

karlhorky:
Open: https://github.com/cheeriojs/cheerio/pull/1845

fb55:
Closing this, as the change seems pretty involved.

PR #1845, opened 2021-05-06
Make type more permissive, allowing browser usage

karlhorky:
Ref: https://github.com/vfile/vfile/blob/f95280cbd0d8d29a3fd49984e91e7c665175bceb/lib/index.js#L13-L17
Ref: https://github.com/cheeriojs/cheerio/issues/1844

### DIFF
--- a/src/load.spec.ts
+++ b/src/load.spec.ts
@@ -35,4 +35,11 @@ describe('.load', () => {
     const $html = cheerio.load(Buffer.from(html));
     expect($html.html()).toBe(html);
   });
+
+  it('(Uint8Array) : should accept a Uint8Array', () => {
+    const html = '<html><head></head><body>foo</body></html>';
+    const enc = new TextEncoder();
+    const $html = cheerio.load(enc.encode(html));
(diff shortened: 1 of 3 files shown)


fb55:
Could you add a test case that doesn't use a buffer? Something like https://github.com/cheeriojs/cheerio/blob/95759d64886d2f4f944daef1d8b706e75aa0154b/src/load.spec.ts#L35

karlhorky:
lol I guess this feature doesn't exist in Cheerio (Uint8Array is not accepted), I guess I'll just close this PR?

5saviahv:
`content` of **Buffer** is later just converted to string (and it is expected to be "utf8"). 

https://github.com/cheeriojs/cheerio/blob/e0c7cbbffa76aa9af6441be10f19431606a1819d/src/parse.ts#L20-L22

you can do the same with **UInt8Array** just use `Buffer.from( content ).toString( 'utf8' )` to convert it to string.

fb55:
Closing this PR as any change seems quite involved. Thanks for proposing this, @karlhorky!

PR #1870, opened 2021-05-15
fix(traversing): Make `filter` work on all collections

### LINKED ISSUE #1867: textNode not passed into .filter method
The `.filter(function)` method only gets DOM elements passed in, while `.each(function)` gets all DOM nodes, including text nodes. This used to work in RC-5, but doesn't in RC-9. And as far as I could tell, this isn't expected behaviour, because I wasn't able to find info indicating that it is in the jQuery docs.

Consider the following html snippet:
```html
<body>
    <div style="color: black;">Hello</div>
    world
    <small>!</small>
</body>
```
If I were to select the body and `.filter` its contents, I won't get the `world` text node in my filter loop. Whereas I will get the `world` text node in my loop if I use `.each` on the contents of the body.

I've created a [repo](https://github.com/jeroentvb/cheerio-filter-bug) demonstrating my findings.

Edit: just checked, it also used to work with RC-6. That means it stopped working with the RC-9 release.

fb55:
Fixes #1867

### DIFF
--- a/src/api/attributes.ts
+++ b/src/api/attributes.ts
@@ -8,6 +8,7 @@ import { text } from '../static';
 import { isTag, domEach, camelCase, cssCase } from '../utils';
 import type { Node, Element } from 'domhandler';
 import type { Cheerio } from '../cheerio';
+import { AcceptedFilters } from '../types';
 const hasOwn = Object.prototype.hasOwnProperty;
 const rspace = /\s+/;
 const dataAttrPrefix = 'data-';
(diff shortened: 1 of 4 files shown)


sehrgut:
This should also iterate over comment nodes. It doesn't appear to do that.

fb55:
@sehrgut This PR removes the restrictions for all set types. With this PR, you can iterate over a `Cheerio<number>` without issue. TS doesn't like this (yet), but this runs in pure JS:

```js
$([1,2,3,4,5]).filter((_,n)=>n%2).toArray()
//=> [1,3,5]
```

sehrgut:
Ah! I see that now. I think I assumed from the new imports list `import { Node, Element, Text, isText } from 'domhandler';` that those would be the only ones picked up by `AcceptedFilters<T>`. Thank you!

PR #1909, opened 2021-06-06
refactor(traversing): Wrap shared behavior

fb55:
Revival of #1731

@5saviahv As this is based on your work, I'd appreciate if you could have a look!


5saviahv:
Nice, I am happy ... sadly I am not very good in this typescript stuff. I try keep up but train is moving way too fast :)

Nice Work

fb55:
> Nice, I am happy ... sadly I am not very good in this typescript stuff. I try keep up but train is moving way too fast :)

No worries, and I can see how TS can be a bit intimidating at first. Happy to help with types, should you feel like cooking up some more PRs :)

PR #1934, opened 2021-06-15
Migrate husky to v6

DavideViolante:
I migrated husky to v6 and it seems to work for me, but review it please.
What I did:
- tried the husky migration tool but without success (I think it needed an husky config file not inside package json)
- installed husky v6 as dev dep
- run `npx husky-init && npm install` as docs states
  - the command created .husky folder with gitignore and an example pre-commit file and add to package json the prepare script
- edited pre-commit file adding npm test, it seems to execute everything to me, but let me know here if the command to execute is different
- tried a commit to check if pre-commit hook is executed, and it is
- bumped version
- to test: prepublishOnly hook? Other hooks?


package.json
```diff
@@ -87,24 +87,15 @@
...
-    "prepublishOnly": "npm run build"
+    "prepublishOnly": "npm run build",
+    "prepare": "husky install"
   },
   "prettier": {
     "singleQuote": true,
     "tabWidth": 2,
     "tsdoc": true
   },
-  "lint-staged": {
```
fb55, line 99: `lint-staged` is still configured in the `package.json`, isn't it?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "cheerio",
-  "version": "1.0.0-rc.10",
```
fb55, line 3: Please don't bump the version as that breaks my release tooling 😄

.husky/pre-commit
```diff
@@ -0,0 +1 @@
+npm test
```
fb55, line 1:
```suggestion
npm x lint-staged
```

fb55 (commented):
Thanks for looking into this! Looks mostly good, but I'd like to keep `lint-staged` around.

fb55:
Thanks so much @DavideViolante, great PR!

PR #2047, opened 2021-08-14
fix(manipulation): Have `text` turn passed values to strings

### LINKED ISSUE #2026: text(1) does not convert a number to a string
In Cheerio 1.0.0-rc.10, when passed a number, the `text()` method does not convert it to a string.

This throws errors if you later try to serialize it to HTML.

(I think this is a regression and the program that I used `text(a_number)` in was working last year).

## Example program

```javascript
const $ = cheerio.load('<p>A program containing <span>a placeholder</span></p>');
$('span').text(1);
const html = $.html();
console.log(html);
```

## Output

```

    str = str.replace(AMP_REGEX, '&amp;').replace(NBSP_REGEX, '&nbsp;');
              ^

TypeError: str.replace is not a function
    at Function.Serializer.escapeString (/Users/david/prog/cheerio-test-case/node_modules/parse5/lib/serializer/index.js:165:15)
    at Serializer._serializeTextNode (/Users/david/prog/cheerio-test-case/node_modules/parse5/lib/serializer/index.js:148:37)
    at Serializer._serializeChildNodes (/Users/david/prog/cheerio-test-case/node_modules/parse5/lib/serializer/index.js:52:26)
    at Serializer._serializeElement (/Users/david/prog/cheerio-test-case/node_modules/parse5/lib/serializer/index.js:93:18)
    at Serializer._serializeChildNodes (/Users/david/prog/cheerio-test-case/node_modules/parse5/lib/serializer/index.js:50:26)
    at Serializer._serializeElement (/Users/david/prog/cheerio-test-case/node_modules/parse5/lib/serializer/index.js:93:18)
    at Serializer._serializeChildNodes (/Users/david/prog/cheerio-test-case/node_modules/parse5/lib/serializer/index.js:50:26)
    at Serializer._serializeElement (/Users/david/prog/cheerio-test-case/node_modules/parse5/lib/serializer/index.js:93:18)
    at Serializer._serializeChildNodes (/Users/david/prog/cheerio-test-case/node_modules/parse5/lib/serializer/index.js:50:26)
    at Serializer._serializeElement (/Users/david/prog/cheerio-test-case/node_modules/parse5/lib/serializer/index.js:93:18)     
```

… because it stores `1` internally and does not convert it to a string. Numbers do not have a `replace`method.

## Expected behaviour

`text(1)` would convert the number to a string and the output would be:

```html
<html><head></head><body><p>A program containing <span>1</span></p></body></html>
```

(It would also be OK if it type checked the input and throw an exception if you tried to set a non-string with `text()`)

## Workaround

Explicitly passing a string works around the issue (this isn't ideal of course, but it serves to prove the source of the problem).

```
$('span').text(`${1}`);
```

fb55:
Fixes #2026

### DIFF
--- a/src/api/manipulation.spec.ts
+++ b/src/api/manipulation.spec.ts
@@ -1901,6 +1901,11 @@ describe('$(...)', () => {
       expect(text).toBe('M&M');
     });
 
+    it('should turn passed values to strings', () => {
+      $('.apple').text(1 as any);
+      expect($('.apple')[0].childNodes[0]).toHaveProperty('data', '1');
+    });
+
     it('( undefined ) : should act as an accessor', () => {
       const $div = $('<div>test</div>');
       expect(typeof $div.text(undefined as any)).toBe('string');
--- a/src/api/manipulation.ts
+++ b/src/api/manipulation.ts
@@ -1004,7 +1004,7 @@ export function text<T extends Node>(
       child.next = child.prev = child.parent = null;
     });
 
-    const textNode = new Text(str);
+    const textNode = new Text(`${str}`);
 
     updateDOM(textNode, el);
   });


dorward:
Great. Thanks for fixing that.

fb55:
@dorward Thanks for the report!

PR #2127, opened 2021-09-28
docs(readme): update "is not a web browser" section

mxschmitt:
Might help the users to understand what the differences are between such projects (browser automation vs. emulated DOM).


fb55:
Thanks a lot @mxschmitt!

PR #2280, opened 2021-12-28
Cheerio breaks some symbols

aivus:
We have upgraded our module from cheerio `0.22` to `v1.0.0-rc.10` and faced with the issue where cheerio breaks some symbols.

We are converting our html to the binary and then pass html to the cheerio. After this we are converting html back.
It was working fine in 0.22, but doesn't work now in latest RC.

I've created a reproduced in this PR. 

@fb55 do you have any ideas why it started to happen?

### DIFF
--- a/src/__tests__/xml.spec.ts
+++ b/src/__tests__/xml.spec.ts
@@ -61,5 +61,17 @@ describe('render', () => {
         '<someelem someattribute="something">hello</someelem>'
       );
     });
+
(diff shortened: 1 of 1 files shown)


fb55:
If you pass UTF-8 encoded strings to Cheerio, things should work as expected. I'm not sure what the output is for the binary format is, but anything but UTF-8 is problematic.

PR #2286, opened 2022-01-02
fix examples: $.html, $.text

milahu:
before

```
> cheerio.html($('.pear'));
Uncaught TypeError: cheerio.html is not a function

> cheerio.text($('body'));
Uncaught TypeError: cheerio.text is not a function
```

after

```
> $.html($('.pear'));
'<li class="pear">Pear</li>'

> $.text($('body'));
'This is content.'
```

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -218,7 +218,7 @@ $.root().html();
 If you want to render the [`outerHTML`](https://developer.mozilla.org/en-US/docs/Web/API/Element/outerHTML) of a selection, you can use the `html` utility functon:
 
 ```js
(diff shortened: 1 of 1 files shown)


Hashbrown777:
This 'fix' is incorrect, the documentation is accurate and the code itself is incorrect (see #2402).
The examples are specifically talking about the static functions.

You've changed the static call `cheerio.text($('body'));` to a call to the member function (of the document) `$.text($('body'));`.
If you're willing to use member functions, you can just do `$('body').text()`.

milahu:
closing for #2402

PR #2506, opened 2022-04-30
fix: Export missing static methods

### LINKED ISSUE #2402: static `text()` method is not actually exported
As mentioned in the README:

 > You may also render the text content of a Cheerio object using the text static method:

```javascript
const $ = cheerio.load('This is <em>content</em>.');
cheerio.text($('body'));
//=> This is content.
```

Yet neither `index.d.ts` nor `index.js` actually forwards `text` from `lib/static.{d.ts,js}` (it appears to re-export methods one-by-one for some reason, and misses `text`).

image

As a workaround I tried importing static directly, but it doesnt like it.

image

Using `import {text} from 'cheerio/lib/static'` gets past typescript complaints, but for some reason doesn't transpile to mjs well.

image

So [at least in typescript targeting ES2020 using modules] it's impossible to use the text static method.

fb55:
Fixes #2402
Closes #2099

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -1,0 +1,10 @@
+import * as cheerio from './index';
+import * as statics from './static';
+
+describe('index', () => {
+  it('should export all static methods', () => {
+    for (const key of Object.keys(statics) as (keyof typeof statics)[]) {
+      expect(cheerio[key]).toBe(statics[key]);
+    }
+  });
+});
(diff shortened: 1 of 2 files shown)


Hashbrown777:
Hm.. I would've done this for #2402 if that's what I thought was appropriate. But from how `src/index.ts` deliberately exports things one-by-one seemed to be what the team wanted (for some reason).

If that's not the case, all the `export const { X } = staticMethods;` lines (basically all past line 90) should be removed, and otherwise if it *is* the case, this fix is inconsistent with how the team wanted things and the missing items should be exported individually.

Having both `export * from './static'` *and* `export {} = staticMethods` is redundant..

fb55:
I left the `export const { X } = staticMethods` calls to propagate deprecations. But I see your point, and made the exports explicit in #2512.

PR #2521, opened 2022-05-03
fix(css): Add malformed fields to previous field

### LINKED ISSUE #1134: .css() splits a property declaration if it contains a semi-colon
Hello !

I am reporting a bug I found.
I am running cheerio `1.0.0-rc.2`
I have an unexpected behavior using the `.css()` function.
It should go along the way to the closing parenthesis of my `url` statement.
What I actually get is that it stops at the semi-colon `;` as if it was the end of my property declaration.

### Initial settings
```js
const cheerio = require('cheerio')
const $ = cheerio.load(`<button style="background-image: url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAC4AAA)"></button>`)
```

### Faulty function
```js
$('button').css('background-image')
```

### Expected output
```js
'url(data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAC4AAA)'
```

### Given output

```js
'url(data:image/png'
```

fb55:
Fixes #1134

### DIFF
--- a/src/api/css.spec.ts
+++ b/src/api/css.spec.ts
@@ -123,6 +123,15 @@ describe('$(...)', () => {
           two: '1',
         });
       });
+
+      it('should add malformed values to previous field (#1134)', () => {
+        const el = cheerio(
+          '<button style="background-image: url(data:image/png;base64,iVBORw0KGgo)"></button>'
+        );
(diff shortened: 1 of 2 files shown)


fb55:
This should fix most issues with the `css` method — until some other case surfaces that we are misparsing 😄

PR #2586, opened 2022-06-10
fixing errors related to enzyme and vitest

BZaragoza:
Trying to setup Vitest and Enzyme in a React project I got this error (https://postimg.cc/HjQyJXGv) related with Cheerio. It seemed a problem related with an export in package.json. I just wrote it and it worked.
Hope it helps you


replaceits:
I get this same error and the above fixes it for me.

PR #2601, opened 2022-06-17
fix(package): Allow imports of `cheerio/lib/utils`

### LINKED ISSUE #2600: It's not possible to import "cheerio/lib/utils" anymore
Enzyme has this import which now fails:
https://github.com/enzymejs/enzyme/blob/master/packages/enzyme/src/Utils.js#L11

blixt:
As used by Enzyme here (and which broke after upgrading Cheerio):

https://github.com/enzymejs/enzyme/blob/master/packages/enzyme/src/Utils.js#L11

Fixes #2600

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,10 @@
     "./lib/slim": {
       "require": "./lib/slim.js",
       "import": "./lib/esm/slim.js"
+    },
+    "./lib/utils": {
+      "require": "./lib/utils.js",
+      "import": "./lib/esm/utils.js"
     }
   },
   "files": [


package.json
```diff
@@ -35,6 +35,10 @@
     "./lib/slim": {
       "require": "./lib/slim.js",
       "import": "./lib/esm/slim.js"
+    },
+    "./lib/utils": {
+      "require": "./lib/utils.js",
```
ljharb, line 40: You’d need a default condition also, to support all the node versions enzyme supports.

blixt: What would it catch? `"exports"` was added in Node.js 12.7 when this should always match the `"require"` case unless using modules.

ljharb: node 13.1-13.7, iirc.

fb55: The other exports here would also fail if that was the case. Let's fix this once it actually becomes an issue for users.

fb55:
Thanks @blixt!

package.json
ljharb: This one will continue to fail for enzyme because that *is* the case; this way we’d have to wait and fix them one by one, which is much more work and much slower.

ljharb:
@fb55 this still doesn't solve the issue, because of https://github.com/cheeriojs/cheerio/pull/2601#discussion_r906825179

fb55:
@ljharb I'm not sure I follow. How can an import not be either an `import` or a `require`? Are there platforms that use export maps for other module systems (UMD etc.)?

ljharb:
@fb55 because there are versions of node - that enzyzme supports - that don't support conditions at all, or that only support the `default` condition.

fb55:
Please correct me if I'm wrong: Doesn't this only affect Node versions that support ESM behind an `--experimental-modules` flag? As (1) nobody should use an experimental flag in production, and (2) there is full support in non-breaking updates, I don't think it is worth to keep supporting this.

ljharb:
No, it does not. It affects node 13.0-13.6, with no flags passed.

ljharb:
To be clear, you're arguing against adding a few bytes to package.json, versus some packages simply not being able to use cheerio at all without a semver-major change.

PR #2610, opened 2022-06-25
Export utils

TerbiumSDK:
Export utils for older webpack 4 to webpack 5 upgrades.


TerbiumSDK:
Push

fb55:
Fixed in https://github.com/cheeriojs/cheerio/pull/2601

Thanks @TerbiumSDK!

PR #2667, opened 2022-07-26
fix grammatical errors

daniijal:


fb55:
Hi @daniijal, thanks for your contribution! These cases are not the most straightforward grammatically, but seem alright to me. Best of luck with your assignment!

PR #2688, opened 2022-08-09
docs: remove example for removed option

seonggukchoi:
I removed a code example about a deleted option.

The `normalizeWhitespace` property was extended from [domhandler](https://github.com/fb55/domhandler), but that package deprecated the property at [21a2f7](https://github.com/fb55/domhandler/commit/21a2f7b6a53c0c61e4c753589c5ba99e611529c9), and removed it at [6503e2](https://github.com/fb55/domhandler/commit/6503e2d22bc1896cfb9553296f761856dc68a27f) finally.

I can't find a replacement, so I just removed the property in README.md.

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -113,11 +113,7 @@ If you need to modify parsing options for XML input, you may pass an extra
 object to `.load()`:
 
 ```js
-const $ = cheerio.load('<ul id="fruits">...</ul>', {
-  xml: {
-    normalizeWhitespace: true,
-  },
-});
+const $ = cheerio.load('<ul id="fruits">...</ul>');
 ```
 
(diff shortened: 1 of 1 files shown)


fb55:
Good find! As the line above speaks about passing options to the `load` function, I do think this should rather be replaced with something like the `withStartIndices` option.

seonggukchoi:
> Good find! As the line above speaks about passing options to the `load` function, I do think this should rather be replaced with something like the `withStartIndices` option.

You're right, I forgot to read the description 😂
I added an optional object using another property at commit [2141ce0](https://github.com/cheeriojs/cheerio/pull/2688/commits/2141ce05980f0cd7e3884034d80d9b4dcaf110b0) instead of removing the object, how do you think about this?

fb55:
Thanks @seonggukchoi!

PR #2811, opened 2022-10-20
fix(traversing): Make `closest` be able to start from text nodes

### LINKED ISSUE #2810: closest() should be able to start from text nodes
A function `closest()` is now working only on tags. It used to work on text nodes too on `1.0.0-rc.10`.

Breaking commit: #2057 

`parent()`, `parents()` and `parentsUntil()` functions still work for text nodes though. I expect `closest()` to behave the same way.

Example:
```js
const $ = cheerio.load('<span>abc</span>');
const textNode = $('span').contents().first();
console.log(textNode.closest('span').length);
```
Expected output is 1 (i.e. closest span has been found), current version returns 0.

I'll create a pull request soon.

Qualtagh:
In `closest`, make one extra step to parent if initial node isn't a tag node.

Fixes #2810


fb55:
Thanks a lot for this PR! The only reason this hasn't been merged yet is that I would want to ensure that no other method has similar issues, and this PR is my reminder. I will get to it eventually.

fb55:
Thanks a lot @Qualtagh

Qualtagh:
Thank you for merging. And thanks a lot for one of the most important NodeJS libraries! You do a great job :muscle:

PR #2826, opened 2022-10-29
added beautiful emojis

Aditya-swarup:


fb55:
Don't think this is an app improvement, sorry 🙂

Issue #2786, opened 2022-10-05
load function is marked deprecated with message saying to use function returned by load


mike-lang:
Hi,

I believe https://github.com/cheeriojs/cheerio/commit/c370f4ed1dac20ed0a2fbe85c1da4d3d3ce383e7 marked the exported load function as deprecated in error.  It looks like the documentation was simply copied from the export of a `Cheerio` constructor over to the load function, including a deprecation message that made sense if someone was using the constructor, but is confusing when `load` is used. 

As an example the following code:
```
let $ = cheerio.load(pageHTML);
```
produces the following JSDoc documentation in code editors:
> The default cheerio instance
@deprecated - Use the function returned by `load` instead.

As best I can tell, this example uses the same method of loading html as is what is shown in the project documentation.

If this was not intentional, could we update the JSDoc to remove this deprecation warning and maybe update the function description?

If it is, could the project documentation we updated to indicated what is now the preferred method of loading and parsing HTML?

fb55:
You are likely importing the default export:

```js
import cheerio from 'cheerio';
```

This will import a pre-loaded instance of cheerio, which can lead to subtle bugs and issues. Instead, try

```js
import * as cheerio from 'cheerio'
```

or

```js
import { load } from 'cheerio'
```

Hope this helps!

fb55:
Re-opening this as the default export should probably be removed.

sthota-fms:
Any progress on this one?

fb55:
See my response above.

sthota-fms:
I keep getting this error when installing this package in my react-native project. My react-native project is pure js and not TS.
<img width="1464" alt="image" src="https://user-images.githubusercontent.com/121507008/220996465-2c1970b5-36da-40e3-8ff3-2865698148e6.png">

It looks like you're missing to export package.json
<img width="453" alt="image" src="https://user-images.githubusercontent.com/121507008/220996343-d85b9f6d-eb73-4a73-95b6-bca21aa8f45b.png">

sthota-fms:
@fb55, Could we please include the above suggestion in the release? I need this package for my solution

fb55:
The default export has been removed.

PR #3022, opened 2023-02-23
Update package.json

sthota-fms:
Need to export package.json for react-native project.

[Please view this issue for further details.](https://github.com/cheeriojs/cheerio/issues/2786)

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,6 +30,7 @@
   "types": "lib/index.d.ts",
   "module": "lib/esm/index.js",
   "exports": {
+    "./package.json": "./package.json",
     ".": {
       "require": "./lib/index.js",
       "import": {


fb55:
Not sure I understand why one would want to import the `package.json` file. Could you elaborate?

sthota-fms:
When I try to install cheerio, I get this warning in my console.

image

Therefore, it's expecting ./package.json.

michaelfaith:
Weighing in here.  It's generally good practice to include an export for the `package.json` so that tooling can use it.  For instance, and app that wants to dynamically display what versions of open source packages its using for legal reporting, can do that easily by loading the `package.json`s of its dependencies.  This seems like an easy thing to add, with no real downside.

fb55:
@michaelfaith added

Issue #581, opened 2014-10-21
.attr(name) should lowercase name property


psayre23:
When looking up an attribute with .attr(), the name of the attribute should be lowercased before looking up in .attribs object.

tlylt:
Hi @fb55, is this issue still relevant? (Happy to take a stab at it if so)

joeheyming:
I can try to take a stab at this.

jennylialiu:
Whitdrawnl to my account wallet

jennylialiu:
(suiing)#581

PR #3306, opened 2023-07-28
Issue #581 - coerce attribute names to lower case

davidfei222:
Fix for issue #581.  Attribute names will now always be coerced to lower case when attempting to get or set attributes to keep the behavior in line with how cheerio.load() handles attribute names.

### DIFF
--- a/src/__fixtures__/fixtures.ts
+++ b/src/__fixtures__/fixtures.ts
@@ -2,7 +2,7 @@ export const fruits = [
   '<ul id="fruits">',
   '<li class="apple">Apple</li>',
   '<li class="orange">Orange</li>',
-  '<li class="pear">Pear</li>',
+  '<li class="pear" FOO="true">Pear</li>',
   '</ul>',
 ].join('');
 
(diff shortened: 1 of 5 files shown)


walshyb:
This PR updates Cheerio's `.attr()` functionality to more closely match jQuery's `.attr()`, which saves and retrieves attributes by lowercasing the input name.

rjpatt:
Adjusted a few existing tests to be able to support tests for this fix.

fb55:
Thanks a lot for this PR! Some notes:

1. The change in the fixture led to a need for quite a few unrelated changes. Please use one of the other fixtures or add a new one.
2. Cheerio has an XML mode for working with XML. XML is case-sensitive, so we don't want to lowercase tag names in XML. You will probably end up with something along the lines of

```js
attrNameToUse = this.options.xmlMode ? attrName : attrName.toLowerCase()
```

davidfei222:
@fb55 Apologies for the late response - my colleagues and I have been very busy at work recently.  The requested changes have been made - please re-review at your earliest convenience.  Thanks!

CyberFlameGO (approved):
LGTM!

Issue #3348, opened 2023-08-27
Incorrect code example in intro documation


VictorPersson:
I was reading the "Basic loading" documentation and pressed the copy button in order to receive the intro code from the [docs](https://cheerio.js.org/docs/basics/loading): 

```js
import * as cheerio from 'cheerio';

const $ = cheerio.load('<html><body><h1>Hello, world!</h1></body></html>');

console.log($('title').text());
// Output: Hello, world!
```

However, the console log turns out empty is empty. This is because it's trying to select and log "title" which does not exist in the provided code: `<html><body><h1>Hello, world!</h1></body></html>`
 
So the output "// Output: Hello, world!" is just ""

fb55:
Fixed!

PR #3349, opened 2023-08-27
Fix old/incorrect selector in intro docs

VictorPersson:
This fixes this issue by adding a class of "title" to the docs: #3348


fb55:
Thanks @VictorPersson! I merged #3421 instead of this one.

PR #3421, opened 2023-10-05
Update loading.md

harishcarpenter1:
import * as cheerio from 'cheerio';

const $ = cheerio.load('<html><body><h1>Hello, world!</h1></body></html>');

console.log($('h1').text());

instead of h1, the title was mentioned so the output was empty string


fb55:
Thanks!

PR #3434, opened 2023-10-14
feat: lowercased lookup with attr/data

joeheyming:
Fix for #581

If you accidentally use a lowercased lookup with .attr() or .data(), we will attempt to retry the lookup with a lowercased name.


fb55:
Thanks! Duplicate of #3306.

PR #3753, opened 2024-04-05
[website/docs] Remove extract.md page because method is not implemented in Cheerio API

nwhitmont:
Cheerio current version as of this PR: 1.0.0-rc.12

**What:**
Remove `website/docs/advanced/extract.md` page from documentation website.

**Context:**
Method `extract()` is not implemented in Cheerio.


fb55:
The docs reflect the repository state, not the npm state. There is a note to that effect at the top 😄

PR #3835, opened 2024-05-28
docs: update example function

iaurg:
The given example function not work like expected, so I made little changes for this to work


fb55:
Thanks for this! I merged https://github.com/cheeriojs/cheerio/pull/3642 instead

Issue #4032, opened 2024-08-16
cheerio 1.0.0 breaks bundling of react native


UNIDY2002:
Minimal example: https://github.com/thu-info-community/thu-info-app/actions/runs/10414518614/job/28843638237?pr=587

```
Error: Unable to resolve module node:stream from /home/runner/work/thu-info-app/thu-info-app/node_modules/cheerio/dist/commonjs/index.js: node:stream could not be found within the project or in these directories:
  ../../node_modules
error Unable to resolve module node:stream from /home/runner/work/thu-info-app/thu-info-app/node_modules/cheerio/dist/commonjs/index.js: node:stream could not be found within the project or in these directories:
  ../../node_modules
  node_modules
  ../../node_modules
  49 | const undici = __importStar(require("undici"));
  50 | const whatwg_mimetype_1 = __importDefault(require("whatwg-mimetype"));
> 51 | const node_stream_1 = require("node:stream");
     |                                ^
  52 | const options_js_1 = require("./options.js");
  53 | const load_parse_js_1 = require("./load-parse.js");
  54 | /**.
  node_modules
  ../../node_modules
  49 | const undici = __importStar(require("undici"));
  50 | const whatwg_mimetype_1 = __importDefault(require("whatwg-mimetype"));
> 51 | const node_stream_1 = require("node:stream");
     |                                ^
  52 | const options_js_1 = require("./options.js");
  53 | const load_parse_js_1 = require("./load-parse.js");
  54 | /**
    at ModuleResolver.resolveDependency (/home/runner/work/thu-info-app/thu-info-app/node_modules/metro/src/node-haste/DependencyGraph/ModuleResolution.js:153:15)
    at DependencyGraph.resolveDependency (/home/runner/work/thu-info-app/thu-info-app/node_modules/metro/src/node-haste/DependencyGraph.js:279:43)
    at /home/runner/work/thu-info-app/thu-info-app/node_modules/metro/src/lib/transformHelpers.js:176:21
    at resolveDependencies (/home/runner/work/thu-info-app/thu-info-app/node_modules/metro/src/DeltaBundler/buildSubgraph.js:56:25)
    at visit (/home/runner/work/thu-info-app/thu-info-app/node_modules/metro/src/DeltaBundler/buildSubgraph.js:107:30)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async Promise.all (index 2)
[...]

PR #4033, opened 2024-08-16
Add a field `browser` to `package.json` root

UNIDY2002:
Fixes https://github.com/cheeriojs/cheerio/issues/4032

Some bundlers (e.g. react-native) requires the field `browser` at `package.json` root to find the correct resolution entry points. (Reference: https://metrobundler.dev/docs/configuration/#resolvermainfields)

This PR adds the correct entry point for browsers to `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,6 +74,7 @@
   },
   "main": "./dist/commonjs/index.js",
   "module": "./dist/esm/index.js",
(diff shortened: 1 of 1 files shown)


fb55:
Thanks a lot @UNIDY2002!

PR #4270, opened 2024-12-01
docs: correct loadBuffer() name in example

arichardsmith:
The example [in the docs](https://cheerio.js.org/docs/api/functions/loadBuffer) uses `cheerio.fromBuffer` instead of `cheerio.loadBuffer`, which confused me for a second when copy-pasting.


fb55:
Good catch, thanks @arichardsmith!

PR #4334, opened 2025-01-04
fix(types): fix ExtractedValue type

ben-tilden:
Noticed this while working with `extract`

TypeScript does not automatically distribute conditionals unless the generic type itself represents a union. So, in `ExtractedValue`, `V['value']` is not distributed unless inferred as a separate generic type.

EDIT: Adding [issue link](https://github.com/cheeriojs/cheerio/issues/4335#issue-2769048688)

```
// Testing distributed conditional over an indexed access type
interface Obj {
  value: string | number;
}

type TestIndexedObj<V extends { value: unknown }> = V['value'] extends number ? 'extends number' : 'does not extend number';
type TestIndexedObjResult = TestIndexedObj<Obj> // "does not extend number"

// Testing distributed conditional over a union
type ObjValue = Obj['value'];

type TestUnion<V> = V extends number ? 'extends number' : 'does not extend number';
type TestUnionResult = TestUnion<ObjValue> // "does not extend number" | "extends number"

// Testing distributed conditional over an inferred generic from an indexed access type
type TestInferredIndexedObj<V extends { value: unknown }> = V['value'] extends infer U ? (U extends number ? 'extends number' : 'does not extend number') : never;
type TestInferredIndexedObjResult = TestInferredIndexedObj<Obj> // "does not extend number" | "extends number"
```

### DIFF
--- a/src/api/extract.spec.ts
+++ b/src/api/extract.spec.ts
@@ -152,6 +152,28 @@ describe('$.extract', () => {
     ).toStrictEqual({ red: 'red=Four' });
   });
 
+  it('should correctly type check custom extraction functions returning non-string values', () => {
+    const $ = load(fixtures.eleven);
+    const $root = $.root();
+
+    expectTypeOf(
+      $root.extract({
+        red: {
+          selector: '.red',
+          value: (el) => $(el).text().length,
+        },
+      }),
+    ).toEqualTypeOf<{ red: number | undefined }>();
+    expect(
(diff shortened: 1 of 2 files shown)


fb55:
Thanks for this awesome PR! Could you add a types test to validate this works (and keeps working)?

ben-tilden:
Apologies for the delay - added!

Issue #4256, opened 2024-11-27
Re-export `Element` type from `domhandler`


demurgos:
Hello,
I am migrating a project from `cheerio@1.0.0-rc.12` using types `"@types/cheerio" @0.22.35` to `cheerio@1.0.0`.

An issue that I hit is that I am not able to name the type returned by the cheerio `$` function: the return type is `Cheerio<Element>`, but the `Element` type comes from the `domhandler` which is not exported.

A workaround would be to add the `domhandler` lib as a direct dependency of my project, but this brittle as it relies on manually keeping both the direct dependency on `domhandler` and the one from `cheerio` in sync. As `domhandler` types are exposed in the public API of Cheerio, they should be re-exported to ensure reliable usage.

`Element` is the most prominent example, but all public types should be reachable from the `cheerio` package. Other examples are `AnyNode`, `Document`, or `ParentNode`.

azerum:
+1, I also suggest to export `AnyNode`. I often need to use `Cheerio<>` with either of those to:

- Define return type of function
- Define parameter of function

PR #4353, opened 2025-01-12
[#4256] Re-export types from domhandler

hexpunk:
Addresses #4256 by re-exporting the types from domhandler that are used by cheerio.


hexpunk:
@jeankhoury0 I don't have the permissions to run the required workflows or merge this.

PR #4454, opened 2025-03-08
docs: use code block for install, mention yarn and bun

Electroid:
Changed the readme to use a code block instead of an inline code tag. Also mentioned yarn and bun as installation methods.


fb55:
Thanks!

PR #4611, opened 2025-06-08
fix bug that get encoded non-latin characters with { xml : true } option

elrion018:


fb55:
Thanks @elrion018! Just pushed https://github.com/cheeriojs/cheerio/commit/55685379074d59d3f259b43f459102bede94e5f6, which documents the existence of the `encodeEntities: "utf8"` option that does what you are looking for!

elrion018:
@fb55  Thank you for quick reply!

But the encoded characters(`'&#x430;&#x431;&#x432;'`) are being obtained still..

스크린샷 2025-06-09 오후 7 18 12

PR #5072, opened 2026-02-13
Fix: Make .attr() remove attributes when value is undefined

### LINKED ISSUE #3307: Can't set existing attribute as undefined
When setting an existing attribute to `undefined` on a Cheerio element, the the attribute does not get removed or set to `undefined`.

However, when setting the same attribute to `null`, the attribute gets removed from the element.

```
const $cheerio = cheerio.load('<ul><li class="pear" foo="bar">Pear</li></ul>');
const $pear = $cheerio('.pear');

$pear.attr('foo', undefined);
const attr1 = $pear.attr('foo');

attr1 === 'bar'           // true
attr1 === undefined // false
$pear.attr('foo')       // [Object: null prototype] { foo: 'bar' }

$pear.attr('foo', null);
const attr2 = $pear.attr('foo');

attr2 === 'bar'           // false
attr2 === undefined  // true
att2 == null                // true
$pear.attr('foo')        // [Object: null prototype] { }
```

veeceey:
## Summary
Fixes #3307 - `.attr('foo', undefined)` now removes the attribute, matching the behavior of `.attr('foo', null)`.

Previously, setting an attribute to `undefined` would not remove it, while setting it to `null` would. This was inconsistent with jQuery's behavior and user expectations.

## Changes
- Updated `setAttr()` function to handle both `null` and `undefined` values
- Fixed `attr()` implementation to use `arguments.length` check instead of `value !== undefined` to properly detect when undefined is explicitly passed
- Updated TypeScript type signatures to include `undefined` in the value parameter
- Added comprehensive tests for undefined attribute removal (both single attribute and map-based updates)

## Test Results
All 129 tests pass, including 3 new tests that verify:
1. Single attribute removal with undefined value
2. Map-based attribute updates with undefined values
3. Chaining behavior when setting attributes to undefined

## Manual Testing
Verified the fix with the exact code from the issue report - both `null` and `undefined` now correctly remove attributes.

### DIFF
--- a/src/api/attributes.spec.ts
+++ b/src/api/attributes.spec.ts
@@ -179,6 +179,25 @@ describe('$(...)', () => {
       expect($pear).toBeInstanceOf($);
     });
 
+    it('(key, value) : should remove attributes when called with undefined value', () => {
+      const $pear = $('.pear').attr('foo', 'bar');
+      expect($pear.attr('foo')).toBe('bar');
+      $pear.attr('foo', undefined);
+      expect($pear.attr('foo')).toBeUndefined();
+    });
+
+    it('(map) : should remove attributes with undefined values', () => {
+      const $pear = $('.pear').attr({
+        foo: 'bar',
+        style: 'color:red',
+      });
+      expect($pear.attr('foo')).toBe('bar');
+      expect($pear.attr('style')).toBe('color:red');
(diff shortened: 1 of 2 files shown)


veeceey:
Friendly ping - any chance someone could take a look at this when they get a chance? Happy to make any changes if needed.

PR #5074, opened 2026-02-15
fix(attributes): preserve non-boolean `hidden` attribute values

### LINKED ISSUE #5073: Attribute value `"until-found"` for hidden attribute is incorrectly normalized
## Description

The `hidden="until-found"` attribute value (introduced in HTML Living Standard) is not correctly preserved when loading HTML via `cheerio.load()`. Instead, it is being normalized to `hidden` or an `undefined`, treating it as a legacy boolean attribute.

https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden#description

```ts
const html = `
  <!DOCTYPE html>
  <!-- Sample: popover -->
  <dialog id="popover-auto" popover="auto"></dialog>
  <dialog id="popover-true" popover></dialog>
  <dialog id="popover-false"></dialog>
  <dialog id="popover-string-empty" popover></dialog>

  <!-- hidden -->
  <div id="hidden-until-found" hidden="until-found"></div>
  <div id="hidden-true" hidden></div>
  <div id="hidden-false"></div>
  <div id="hidden-string-empty" hidden></div>
`;
```

```bash
~/projects/stackblitz-starters-bnzemr2z
❯ node index.js
{ id: 'popover-auto', expected: 'auto', actual: 'auto' }
{ id: 'popover-true', expected: '', actual: '' }
{ id: 'popover-false', expected: undefined, actual: undefined }
{ id: 'popover-string-empty', expected: '', actual: '' }
{ id: 'hidden-until-found', expected: 'until-found', actual: 'hidden' }
{ id: 'hidden-true', expected: '', actual: 'hidden' }
{ id: 'hidden-false', expected: undefined, actual: undefined }
{ id: 'hidden-string-empty', expected: '', actual: 'hidden' }
``` 
https://stackblitz.com/edit/stackblitz-starters-bnzemr2z

## Environment
cheerio version: 1.2.0
node version: v24.1.0

rururux:
fixes: #5073

## Changes

Previously, the `.attr()` method could not correctly retrieve the newly implemented `until-found` value for the `hidden` attribute. It would instead return `"hidden"` or `undefined` due to being treated strictly as a boolean attribute.
In this fix, I have removed `hidden` from the `rboolean` regex/variable to allow it to be handled as an enumerated attribute.

## Testing

I have added a test case to ensure that the value of the hidden attribute can be correctly read from elements like `<div hidden="until-found">`.


fb55 (commented):
Thanks @rururux! Could we still return a boolean if it is hidden, and extend the return type to "until-found"? Otherwise this will break old code.

PR #5120, opened 2026-03-03
fix: improve error messages with helpful suggestions

theluckystrike:
## Summary

This PR improves error messages in cheerio to help developers debug faster by providing actionable context.

### `cheerio.load()` error
- Shows the actual type received (null, undefined, etc.) instead of a generic message
- Lists valid input types (string, Buffer, DOM node)
- Includes a usage example and debugging hint

### Selector type error
- Explains what types are accepted, with examples

## Example Improved Error Messages

Before:
```
Error: cheerio.load() expects a string
```

After:
```
Error: cheerio.load() expects a string, Buffer, or DOM node as content. Received: undefined. Hint: If you're passing a variable, make sure it's not undefined or null. Example: cheerio.load('<html>...</html>')
```

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -133,7 +133,13 @@ export function getLoad(
     isDocument = true,
   ): CheerioAPI {
     if ((content as string | null) == null) {
-      throw new Error('cheerio.load() expects a string');
+      const actualType = content === null ? 'null' : typeof content;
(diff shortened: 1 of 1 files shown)


theluckystrike:
Closing — this needs more review. Thanks!

PR #5256, opened 2026-05-16
fix(attributes): lowercase attribute name before lookup in getAttr and setAttr

chu0119:
## Summary
Fixed `.attr()` to handle case-insensitive attribute name lookup, matching HTML specification and jQuery behavior.

## Why
HTML attribute names are case-insensitive per the spec. htmlparser2 stores all attribute names in lowercase within the internal `attribs` object. When a user passes an uppercase attribute name (e.g., `$el.attr('CLASS')`), the lookup fails because the stored key is `class`.

## What changed
- `getAttr`: Lowercase the attribute name before looking it up in `elem.attribs`, and in the option/checkbox value fallback comparisons
- `setAttr`: Lowercase the attribute name before storing, maintaining consistency with htmlparser2's lowercase storage
- Added test case for case-insensitive attribute lookup

## Test plan
- All 128 existing tests pass
- New test `valid attr should be case-insensitive for HTML documents` added

### DIFF
--- a/src/api/attributes.spec.ts
+++ b/src/api/attributes.spec.ts
@@ -39,6 +39,11 @@ describe('$(...)', () => {
       expect(cls).toBe('apple');
     });
 
+    it('(valid key) : valid attr should be case-insensitive for HTML documents', () => {
+      const cls = $('.apple').attr('CLASS');
+      expect(cls).toBe('apple');
+    });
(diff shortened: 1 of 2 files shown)


fb55:
Thanks! Duplicate of #3306.

PR #5260, opened 2026-05-20
fix(attributes): make removeAttr case-insensitive for HTML documents

### LINKED ISSUE #5257: Bug: .removeAttr() should be case-insensitive for HTML documents
## Describe the bug

## Steps to reproduce

$("div").removeAttr("CLASS");

## Expected behavior

## Environment

algojogacor:
This is an independent contribution made by an individual developer. This work is not associated with any hackathon, competition, or coordinated PR campaign.

## Summary

Fixes #5257

`.removeAttr()` fails to remove attributes when called with an uppercase name (e.g., `.removeAttr('CLASS')`) because htmlparser2 stores HTML attribute names in lowercase per the HTML spec, but the `removeAttribute` helper does a case-sensitive lookup.

## Root Cause

The `removeAttribute` function at `src/api/attributes.ts:840` performs `Object.hasOwn(elem.attribs, name)` without normalizing the attribute name. htmlparser2 lowercases all attribute names when parsing HTML documents, so `elem.attribs['class']` exists but `elem.attribs['CLASS']` does not. The same issue affects the `delete` on the following line.

## Fix

Normalize the attribute name to lowercase with `toLowerCase()` before checking existence and deleting. This matches htmlparser2's storage convention for HTML documents.

## Changes

- `src/api/attributes.ts`: normalize attribute name in `removeAttribute` (`+3 -1` lines)

## Testing

- Existing test suite: 127/127 passing (vitest) ✅
- `.removeAttr('class')` on element with lowercase `class` attribute: still works ✅
- `.removeAttr('CLASS')` on element with lowercase `class` attribute: now correctly removes it ✅
- `.removeAttr('id class')` space-separated: both attributes removed ✅

### DIFF
--- a/src/api/attributes.ts
+++ b/src/api/attributes.ts
@@ -838,9 +838,12 @@ export function val<T extends AnyNode>(
  * @param name - Name of the attribute to remove.
  */
 function removeAttribute(elem: Element, name: string) {
-  if (!(elem.attribs && Object.hasOwn(elem.attribs, name))) return;
+  // Normalize attribute name to lowercase; htmlparser2 stores HTML attribute
+  // names in lowercase per the HTML spec (case-insensitive attributes).
+  const normalizedName = name.toLowerCase();
+  if (!(elem.attribs && Object.hasOwn(elem.attribs, normalizedName))) return;
 
-  delete elem.attribs[name];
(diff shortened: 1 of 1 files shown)


fb55:
Thanks! Duplicate of #5290.

PR #5266, opened 2026-05-21
fix: support root :scope selectors

### LINKED ISSUE #4509: :scope selector not working on root node
To select the direct child of the root node, I'm using :scope pseudo class in the selector.
For example,
```
const html=`
<collection>
  <id value="a"/>
  <relatedCollections>
    <collection>
      <id value="a1"/>
    </collection>
  </relatedCollections>
</collection>
<collection>
  <id value="b"/>
  <relatedCollections>
    <collection>
      <id value="b1"/>
    </collection>
  </relatedCollections>
</collection>
`
let $ = cheerio.load(html,{xmlMode:true},false);
```
to get the root collection id, I'm using 
```
  $(":scope> collection > id").map(function(){ return $(this).attr("value");}).get()
```
, which should return ["a","b"]
however, it is currently returning empty result []

Is it a bug? Any solution? or any workaround?

Thanks and Best Regards,
Shady

puneetdixit200:
Fixes #4509.

This updates root-scoped selector handling so `:scope > ...` works when the current Cheerio selection is the document root. The document root cannot be matched directly by the underlying element-only selector path, so the root `:scope` case is rewritten to select from top-level root elements while keeping the existing element-scoped path unchanged.

I added a regression test that includes a nested matching element to make sure `:scope > ...` only matches direct children of the document root.

Validated with:
- `npm run test:vi -- src/api/traversing.spec.ts -t ":scope on the document root"`
- `npm run test:vi -- src/api/traversing.spec.ts`
- `npm run test:vi`
- `npm run lint:ts`
- `npm run lint:biome`
- `npx eslint src/api/traversing.ts src/api/traversing.spec.ts`
- `npm run build`

Note: I also tried `npm run test`; it currently stops during repo-wide ESLint while linting `website/src/content.config.ts` because `astro/tsconfigs/strict` is not available from the root install.

### DIFF
--- a/src/api/traversing.spec.ts
+++ b/src/api/traversing.spec.ts
@@ -96,6 +96,19 @@ describe('$(...)', () => {
       expect(q('.a').find(':scope')).toHaveLength(1);
     });
 
+    it('should support :scope on the document root', () => {
+      const q = load(
+        '<wrapper><collection><id value="nested"/></collection></wrapper><collection><id value="a"/></collection><collection><id value="b"/></collection>',
+        { xml: true },
+        false,
+      );
+      expect(
+        q(':scope > collection > id')
+          .map((_, el) => q(el).attr('value'))
+          .get(),
+      ).toStrictEqual(['a', 'b']);
+    });
+
     it('should query case-sensitively when in xml mode', () => {
       const q = load('<caseSenSitive allTheWay>', { xml: true });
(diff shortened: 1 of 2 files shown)


fb55:
Hi @puneetdixit200, this change really should be split across the dependencies, not implemented exclusively in Cheerio.

PR #5304, opened 2026-06-14
Fix links to Cheerio methods

yanqui-uxo:


fb55:
Thanks! Duplicate of #5344, which fixes this in the link-generation step rather than per link.

PR #5313, opened 2026-06-17
fix: removeAttr case-insensitive for HTML documents

rafaumeu:
## Problem (#5257)

`.removeAttr()` did not lowercase the attribute name before removing it from `elem.attribs`. Since htmlparser2 stores all attribute names in lowercase (per HTML spec), calling `.removeAttr("CLASS")` would fail to remove a `class` attribute.

```js
const $ = cheerio.load('<div class="test"></div>');
$('div').removeAttr('CLASS');
console.log($('div').attr('class')); // Should be undefined, returns "test"
```

## Fix

Lowercase the attribute name when not in `xmlMode`, matching the behavior of `attr()` and the HTML spec. XML mode preserves case sensitivity.

## Testing

- All 129 existing attribute tests pass
- Added 2 new tests: case-insensitive removal in HTML mode, case-sensitive removal in XML mode
- Biome lint passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
         "default": "./dist/commonjs/index.js"
       }
     },
-    "./package.json": "./package.json",
     "./slim": {
       "import": {
         "browser": {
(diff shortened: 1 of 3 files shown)


fb55:
Thanks! Duplicate of #5290.

PR #5314, opened 2026-06-17
fix: boolean attributes with values return actual value (#5073)

rafaumeu:
## Problem (#5073)

When an HTML boolean attribute has a value (e.g. `hidden="until-found"`), `.attr()` incorrectly returned the attribute name (`"hidden"`) instead of the actual value (`"until-found"`).

This happened because `getAttr()` checked if the attribute name matched the `rboolean` regex and unconditionally returned the attribute name for all boolean attributes, ignoring the actual stored value.

## Fix

Modified `getAttr()` to only return the attribute name for boolean attributes when the stored value is an empty string (pure boolean attribute). For attributes with actual values, return the value.

## Behavior

| HTML | Before | After |
|------|--------|-------|
| `<div hidden="until-found">` | `"hidden"` | `"until-found"` |
| `<div hidden>` | `"hidden"` | `"hidden"` |
| `<div hidden="true">` | `"hidden"` | `"true"` |
| `<div hidden="false">` | `"hidden"` | `"false"` |
| `<div hidden="">` | `"hidden"` | `"hidden"` |

## Testing

- All 136 existing attribute tests pass
- Added 8 new tests covering boolean attributes with various values
- Biome lint passes

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,7 +43,6 @@
         "default": "./dist/commonjs/index.js"
       }
     },
-    "./package.json": "./package.json",
     "./slim": {
       "import": {
         "browser": {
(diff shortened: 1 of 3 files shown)


fb55:
Thanks! Duplicate of #5074.

PR #5315, opened 2026-06-18
docs: add Deno to install instructions

bartlomieju:
👋 Bartek from the Deno team here. Thanks for cheerio, it's a go-to for HTML scraping on Deno too.

The README's Installation section shows npm and bun. Since Deno is a drop-in replacement for npm these days, this adds a Deno line in parity:

```sh
deno install cheerio
```

(As of Deno 2.8, no `npm:` prefix needed, same shape as `npm install`.)

Totally fine to close this if you'd rather keep the list short. Happy to tweak.

_Disclosure: I work on Deno, so I have an interest here. The goal is parity with the package managers you already list, not promotion. This PR was prepared with AI assistance under human supervision: I review every change myself and personally respond to any comments or review feedback._


fb55:
Thanks! Ooc: I've also been slowly adding Cheerio's dependencies to JSR, is that still worthwhile? The removal of the prefix sounds like JSR is deprecated?

bartlomieju:
> Thanks! Ooc: I've also been slowly adding Cheerio's dependencies to JSR, is that still worthwhile? The removal of the prefix sounds like JSR is deprecated?

@fb55 it's not deprecated, but people were complaining that npm being more popular requires 4 additional characters, that's why we removed the prefix.

PR #5321, opened 2026-06-23
fix: lowercase attribute name in attr() getter and setter for HTML mode

### LINKED ISSUE #581: .attr(name) should lowercase name property

syu-toutousai:
HTML attribute names are case-insensitive. When looking up or setting attributes via `.attr()`, the name should be lowercased before reading/writing to `.attribs` in non-XML (HTML) mode.

Lowercase the `name` parameter in `getAttr`, `setAttr`, and `removeAttr` when `xmlMode` is false.

Closes #581

PayPal: n6085530@gmail.com

### DIFF
--- a/src/api/attributes.ts
+++ b/src/api/attributes.ts
@@ -61,6 +61,11 @@ function getAttr(
     return elem.attribs;
   }
 
+  // HTML attribute names are case-insensitive
+  if (!xmlMode) {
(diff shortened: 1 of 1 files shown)


fb55:
Thanks! Duplicate of #3306.

PR #5344, opened 2026-07-09
fix(website): lowercase API doc links for case-sensitive hosting

### LINKED ISSUE #5303: Incorrect links on API documentation pages
On any page under /docs/api, any link with uppercase letters leads to a 404 page. The correct URLs are in lowercase.

deveshkharve:
## Issue
TypeDoc-generated API docs link to mixed-case paths such as `/docs/api/interfaces/CheerioAPI`, but Astro serves lowercase routes at `/docs/api/interfaces/cheerioapi/`. On case-sensitive static hosting, those internal links returns with 404.

```md
[`CheerioRequestOptions`](/docs/api/interfaces/CheerioRequestOptions.md)
```

## Repro

Open https://cheerio.js.org/docs/api/functions/fromurl and click `CheerioRequestOptions` in the parameters table — the link goes to `/docs/api/interfaces/CheerioRequestOptions` and returns **404**.

```bash
curl -sI https://cheerio.js.org/docs/api/interfaces/CheerioRequestOptions | head -1
# HTTP/2 404

curl -sI https://cheerio.js.org/docs/api/interfaces/cheerioapi | head -1
# HTTP/2 301
```

The same links often work locally on macOS because the filesystem is case-insensitive, which hides the bug during `astro preview`.

## Root cause

TypeDoc preserves TypeScript symbol casing in generated markdown links. Astro's content collection lowercases slugs when building static routes. The existing remark plugin only stripped `.md` extensions and did not lowercase paths. It also runs before GFM table parsing, so links inside TypeDoc parameter tables were left unchanged.

## Fix

Extract a shared normalizer and apply it in both remark and rehype stages. Lowercase `/docs/api/` paths and strip `.md` extensions while preserving hash fragments.

```ts
export function normalizeApiDocLink(url: string): string | null {
  if (!url.startsWith('/docs/api/')) {
    return null;
  }

  const [path, hash] = url.split('#');
  const normalizedPath = path.replace(/\.md$/, '').toLowerCase();
  return hash ? `${normalizedPath}#${hash}` : normalizedPath;
}
```

The remark plugin fixes guide pages and standard markdown links. The rehype plugin fixes links rendered from TypeDoc GFM tables.

Fixes #5303

## Tests

- `cd website && npm run build`
- `grep -roh 'href="/docs/api/[^"]*[A-Z]' dist --include='*.html' | wc -l` → `0`
- Verified `/docs/api/functions/fromurl` links resolve (e.g. `CheerioRequestOptions`, `CheerioAPI`)
- Verified guide page links on `/docs/basics/traversing` resolve (e.g. `/docs/api/classes/cheerio#find`)

### DIFF
--- a/website/astro.config.mjs
+++ b/website/astro.config.mjs
@@ -5,6 +5,7 @@ import tailwindcss from '@tailwindcss/vite';
 import { defineConfig } from 'astro/config';
 import remarkDirective from 'remark-directive';
 import { rehypeExternalLinks } from './src/plugins/rehype-external-links.ts';
+import { rehypeFixTypedocLinks } from './src/plugins/rehype-fix-typedoc-links.ts';
 import { remarkAdmonitions } from './src/plugins/remark-admonitions.ts';
 import { remarkFixTypedocLinks } from './src/plugins/remark-fix-typedoc-links.ts';
 import { remarkLiveCode } from './src/plugins/remark-live-code.ts';
@@ -30,6 +31,6 @@ export default defineConfig({
       remarkFixTypedocLinks,
       remarkLiveCode,
     ],
-    rehypePlugins: [rehypeExternalLinks],
+    rehypePlugins: [rehypeFixTypedocLinks, rehypeExternalLinks],
   },
 });
--- a/website/src/plugins/normalize-api-doc-link.ts
+++ b/website/src/plugins/normalize-api-doc-link.ts
@@ -1,0 +1,12 @@
+const markdownExtensionRe = /\.md$/;
+
+/** Normalize internal API doc URLs for case-sensitive static hosting. */
+export function normalizeApiDocLink(url: string): string | null {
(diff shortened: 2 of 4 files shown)


fb55:
Thanks for the thorough diagnosis here. This has since been fixed by #5388, which replaced the plugin this PR patches with `remarkInternalLinks`. That lowercases `/docs/api/` paths and also handles trailing slashes and asset paths, so the normalizer here is now redundant.

The GFM table links are covered too. TypeDoc emits markdown tables rather than HTML, so those links are ordinary mdast `link` nodes that the remark pass already reaches — the separate rehype stage isn't needed. Verified on the deployed site: `/docs/api/functions/fromurl/` no longer emits any mixed-case `/docs/api/` hrefs, and `/docs/api/interfaces/cheeriorequestoptions/` resolves.

Closing as fixed — sorry this sat long enough to be overtaken.

PR #5385, opened 2026-07-30
fix(attributes): resolve prop URLs the way the DOM does

### LINKED ISSUE #5384: prop('href')/prop('src') ignore <base href> and skip area/base/script/embed/track/input
## Describe the bug

`prop('href')` and `prop('src')` diverge from the DOM in two ways:

1. **An in-document `<base href>` is ignored.** The HTML spec resolves URL-reflecting
   attributes against the [document base URL](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#document-base-url),
   which is the first `<base href>` in the document when present, and the document URL
   otherwise. Cheerio always resolves against `options.baseURI`, so every scraped URL on a
   page that uses `<base href>` comes out pointing at the wrong path.

2. **Only 6 of the 13 elements that reflect these attributes are resolved.** Today `href` is
   resolved for `a` and `link`, and `src` for `img`, `iframe`, `audio`, `video` and `source`
   ([attributes.ts](https://github.com/cheeriojs/cheerio/blob/main/src/api/attributes.ts)).
   The DOM also reflects `href` on `area` and `base`, and `src` on `script`, `embed`, `track`
   and `input`. For those elements `prop()` silently returns the raw relative attribute
   instead of a resolved URL, which is easy to miss because the same call works for `<a>`.

## Steps to reproduce

```js
import { load } from 'cheerio';

const html = `
  <head><base href="/sub/dir/"></head>
  <body>
    <a href="p.html">a</a>
    <area href="p.html">
    <script src="j.js"></script>
  </body>`;

const $ = load(html, { baseURI: 'https://example.com/base/page.html' });

$('a').prop('href');      // 'https://example.com/base/p.html'   <- ignores <base>
$('area').prop('href');   // 'p.html'                            <- not resolved at all
$('script').prop('src');  // 'j.js'                              <- not resolved at all
```

Same document in jsdom (which follows the spec here):

```js
import { JSDOM } from 'jsdom';

const d = new JSDOM(html, { url: 'https://example.com/base/page.html' }).window.document;

d.querySelector('a').href;     // 'https://example.com/sub/dir/p.html'
d.querySelector('area').href;  // 'https://example.com/sub/dir/p.html'
d.querySelector('script').src; // 'https://example.com/sub/dir/j.js'
```

## Expected behavior

`prop('href')` / `prop('src')` should match the DOM:

- resolve against the first `<base href>` in the document when there is one, falling back to
  `options.baseURI` (a `<base>`'s own `href` resolves against the document URL, not itself);
- resolve `href` for `a`, `area`, `base`, `link` and `src` for `audio`, `embed`, `iframe`,
  `img`, `input`, `script`, `source`, `track`, `video`;
- keep ignoring `<base>` in XML mode, where the element carries no HTML semantics.

Invalid or missing `<base href>` values should be ignored rather than throwing, and the first
`<base href>` wins when a document contains several, as in the spec.

## Environment

- cheerio version: `main` (f8f3393) and 1.2.0
- Node.js: 26
- verified against jsdom 28 for DOM parity

Happy to open a PR — I have the fix and tests ready, following CONTRIBUTING (issue first).

vojtisprime11:
Fixes #5384.

`prop('href')` and `prop('src')` diverged from the DOM in two ways. Both are fixed here, with
behaviour checked against jsdom for every element in the matrix.

### 1. An in-document `<base href>` was ignored

URL-reflecting attributes resolve against the
[document base URL](https://html.spec.whatwg.org/multipage/urls-and-fetching.html#document-base-url),
which is the first `<base href>` when the document has one. Cheerio always used `options.baseURI`,
so on a page that sets a base every scraped URL pointed at the wrong path.

```js
const $ = load('<head><base href="/sub/dir/"></head><a href="p.html">a</a>', {
  baseURI: 'https://example.com/base/page.html',
});

$('a').prop('href');
// before: https://example.com/base/p.html
// after:  https://example.com/sub/dir/p.html   (matches jsdom and browsers)
```

Details, all covered by tests:

- the first `<base href>` in the document wins, as in the spec;
- a `<base>`'s own `href` resolves against the document URL, not against itself;
- a `<base>` with no `href`, or an unparseable one, falls back to `options.baseURI` instead of
  throwing;
- `<base>` carries no HTML semantics in XML mode, so it is ignored there.

The lookup only walks the children of `<head>`, where the element is required to live, so
documents without a `<base>` pay a constant cost rather than a tree scan per `prop()` call.

### 2. Only 6 of the 13 elements that reflect these attributes were resolved

`href` was resolved for `a` and `link`, `src` for `img`, `iframe`, `audio`, `video` and `source`.
The DOM also reflects `href` on `area` and `base`, and `src` on `script`, `embed`, `track` and
`input`. For those, `prop()` returned the raw relative attribute — easy to miss, because the same
call on an `<a>` returns an absolute URL.

```js
$('script').prop('src'); // before: 'j.js'   after: 'https://example.com/sub/dir/j.js'
```

The element/attribute pairs now live in one `urlAttributeElements` map instead of a chained
boolean condition, so the supported set is readable and easy to extend.

### Not included

`form[action]`, `blockquote[cite]`, `q[cite]`, `ins`/`del[cite]` and `video[poster]` are also
URL-reflecting in the DOM, but `prop()` has never resolved them and adding new prop names felt
like separate scope. Happy to follow up if you want them.

### Checks

- `vitest run`: 800 passing (7 new tests), no type errors
- `eslint src`, `biome check src`, `tsc --noEmit`: clean
- docs: added the `<base href>` precedence note to the extract guide, and updated the `prop`
  JSDoc with the full element list

(`eslint .` fails on this machine for `website/astro.config.mjs` because the website workspace
dependencies aren't installed — unrelated to this change.)


vojtisprime11:
Thanks — both review findings were correct, and the second one was worse than reported. Fixed in 4d51212.

**1. An href-less `<base>` shadowed a later valid one.** The spec takes the first base element *that has an href*, so `<base><base href="/assets/">` must use `/assets/`. The lookup now skips href-less elements. jsdom agrees:

```js
new JSDOM('<head><base><base href="/assets/"></head><a href="p.html">a</a>',
  { url: 'https://example.com/page/1' }).window.document.querySelector('a').href;
// 'https://example.com/assets/p.html'
```

**2. `data:`/`javascript:` bases didn't just resolve oddly — they threw.** Resolving a relative URL against an opaque-path base is a `TypeError`, so `prop('href')` crashed on any page carrying such a base:

```
before: <head><base href="data:text/html,x"></head><a href="p.html">
        $('a').prop('href')  ->  TypeError: Invalid URL
after:  ->  'http://example.com/page/p.html'
```

Those schemes are also what [Is base allowed for Document?](https://html.spec.whatwg.org/multipage/semantics.html#is-base-allowed-for-document) blocks, so falling back to the document URL is both spec-correct and crash-free. Worth noting jsdom does *not* implement that check — it sets `document.baseURI` to the `data:` URL and then returns the raw attribute from `a.href` — so this one is spec/browser behaviour rather than jsdom parity.

**3. Docs wording** narrowed to HTML mode, with XML mode called out explicitly.

New tests: href-less base followed by a valid one, `data:` base, `javascript:` base. Suite is at 802 passing, no type errors; `eslint src` and `biome check src` clean. Verified the new tests fail against the previous commit (wrong URL for case 1, `TypeError` for case 2).

I left the "first base wins" behaviour for an *invalid* first href as a fallback to the document URL rather than continuing to later bases, matching the spec's frozen-base-URL handling.

Issue #5368, opened 2026-07-25
fromURL doesn't seem to automatically set the document's URL as expected


tomjbradley:
On [this page](https://cheerio.js.org/docs/advanced/extract/#:~:text=As%20an%20attribute%20with%20special%20logic%20inside%20the%20prop%20method%2C%20hrefs%20will%20be%20resolved%20relative%20to%20the%20document%E2%80%99s%20URL%2E%20The%20document%E2%80%99s%20URL%20will%20be%20set%20automatically%20when%20using%20fromURL%20to%20load%20the%20document%2E%20Otherwise%2C%20use%20the%20baseURI%20option%20to%20specify%20the%20documents%20URL%2E) it mentions that the document URL, for the purpose of resolving relative hrefs, should be set automatically when using fromURL, however this doesn't appear to be the case.

For instance, the following code outputs "/docs/intro" instead of "https://cheerio.js.org/docs/intro":
```js
const cheerio = require("cheerio");

(async () => {
    const $ = await cheerio.fromURL("https://cheerio.js.org/");

    const href = $("a:contains(Learn)").prop("href");

    console.log(href);
})();
```

vojtisprime11:
Root cause, for the record: this is fixed on `main`, but only by accident, and it is still broken in the released 1.2.0.

In 1.2.0, `fromURL` derives the base URL like this:

```js
const history = res.context?.history;
const baseURI = history ? history[history.length - 1] : urlObject;
```

Because the redirect interceptor is always composed, `res.context.history` **exists but is empty** when no redirect happened, so `history[history.length - 1]` is `undefined` and `baseURI` ends up `undefined` — `prop('href')` then returns the raw attribute. With a redirect there is an entry in `history`, which is why the existing `should follow redirects` test passes and the no-redirect path went unnoticed.

On `main` the same line reads:

```js
const baseURI = history?.at(-1) ?? urlObject;
```

which falls back to the requested URL, so the reported case works there:

```
main (f8f3393):  await cheerio.fromURL('https://cheerio.js.org/') -> $('a:contains(Learn)').prop('href')
  => 'https://cheerio.js.org/docs/intro'
cheerio 1.2.0:
  => '/docs/intro'
```

[...]

PR #5386, opened 2026-07-30
test(fromURL): cover `baseURI` when no redirect happens

vojtisprime11:
Refs #5368. Test-only change.

`fromURL` always composes undici's redirect interceptor, so `res.context.history` exists but is **empty** when the response was not redirected. The released 1.2.0 reads the final URL as `history[history.length - 1]`, which is `undefined` in that case, so `baseURI` is never set and `prop('href')` returns the raw attribute — exactly what #5368 reports.

`main` is not affected, because the expression became `history?.at(-1) ?? urlObject` during an unrelated refactor (46bfdf3). That means the fix was incidental and nothing pins it down: the existing `should follow redirects` test only exercises the path where `history` has an entry.

This adds a test for the no-redirect path, covering a document-relative and a root-relative href. Verified both ways:

- against `main`: passes
- against the 1.2.0 expression (`history ? history[history.length - 1] : urlObject`): fails with `expected 'link' to be 'http://localhost:PORT/dir/link'`

`vitest run src/index.spec.ts` -> 14 passing, no type errors; `biome check` and `eslint` clean on the touched file.

### DIFF
--- a/src/index.spec.ts
+++ b/src/index.spec.ts
@@ -221,4 +221,18 @@ describe('fromURL', () => {
     );
     expect($('a').prop('href')).toBe(`http://localhost:${port}/final/link`);
   });
+
+  it('should set `baseURI` to the requested URL when there is no redirect', async () => {
+    const port = await createTestServer(
+      'text/html',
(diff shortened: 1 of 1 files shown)


fb55:
Thanks @vojtisprime11!

PR #5389, opened 2026-07-30
fix: make removeAttr case-insensitive for HTML, preserve hidden=until-found

### LINKED ISSUE #5073: Attribute value `"until-found"` for hidden attribute is incorrectly normalized

### LINKED ISSUE #5257: Bug: .removeAttr() should be case-insensitive for HTML documents

marceli1404:
## Description

Makes HTML attribute removal case-insensitive for consistency with the HTML spec, and preserves the special `hidden="until-found"` value when reading attributes.

### Changes

- **`.removeAttr()` case-insensitive in HTML mode**: Lowercases attribute names before removal in HTML documents, matching how htmlparser2 stores attribute names. XML mode preserves case-sensitive removal.
- **Hidden attribute value preservation**: `.attr('hidden')` now returns `"until-found"` when set, instead of normalizing to `true`. This aligns with the [HTML Living Standard](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/hidden#description).

### Tests

- Added test for hidden attribute value (lowercase and mixed-case)
- Added tests for case-insensitive HTML removeAttr
- Added test confirming XML removeAttr remains case-sensitive

Closes #5257
Closes #5073

### DIFF
--- a/src/api/attributes.spec.ts
+++ b/src/api/attributes.spec.ts
@@ -44,6 +44,11 @@ describe('$(...)', () => {
       expect(attr).toBe('autofocus');
     });
 
+    it('(valid key) : should preserve hidden="until-found"', () => {
+      const $el = cheerio('<div hidden="until-found">');
+      expect($el.attr('hidden')).toBe('until-found');
+    });
+
     it('(key, value) : should set one attr', () => {
       const $pear = $('.pear').attr('id', 'pear');
       expect($('#pear')).toHaveLength(1);
@@ -796,6 +801,12 @@ describe('$(...)', () => {
       $body.removeAttr('class');
 
       expect($text('body').html()).toBe(mixedText);
+    });
+
+    it('(key) : should be case-insensitive for HTML', () => {
+      const $apple = $('.apple');
+      $apple.removeAttr('CLASS');
+      expect($apple.attr('class')).toBeUndefined();
     });
   });
 
--- a/src/api/attributes.spec.ts
+++ b/src/api/attributes.spec.ts
@@ -819,6 +819,7 @@ describe('$(...)', () => {
         xmlMode: true,
       });
       $xml('foo').removeAttr('BAR');
+      expect($xml('foo').attr('BAR')).toBeUndefined();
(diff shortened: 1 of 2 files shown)


fb55:
Thanks! Duplicate of #5290 (removeAttr) and #5074 (hidden=until-found).

Issue #687, opened 2015-03-19
.replaceWith() not affecting all elements


petemill:
Using v.0.18 and v.0.13, on a cheerio object with multiple elements, calling .replaceWith() does not affect all elements.

For example the following test script:

``` javascript
var cheerio = require('cheerio');
console.log(cheerio.version);
var html = '<h1><div>one</div></h1><h2>two</h2><div><h1><div>three</div></h1><h2>four</h2><h1>five</h1><h1>six</h1></div>';
console.log(html)
var $ = cheerio.load(html);

console.log($('h1').length);
$('h1').each(function(i, elem){
    $(elem).replaceWith($('<h3>Replaced</h3>'));
});
console.log($('h3').length);
console.log($.html());


var $ = cheerio.load(html);

console.log($('h1').length);
$('h1').replaceWith($('<h3>Replaced</h3>'));
console.log($('h3').length);
console.log($.html());
```

Output:

```
0.18.0
<h1><div>one</div></h1><h2>two</h2><div><h1><div>three</div></h1><h2>four</h2><h1>five</h1><h1>six</h1></div>
4
4
<h3>Replaced</h3><h2>two</h2><div><h3>Replaced</h3><h2>four</h2><h3>Replaced</h3><h3>Replaced</h3></div>
4
1
<h2>two</h2><div><h2>four</h2><h3>Replaced</h3></div>
```

As can be seen, iterating through each element and calling .replaceWith on each element works as expected, but calling .replaceWith on a cheerio object with all 4 matched elements, only 1 gets replaced. Weirdly enough, in cheerio v0.13, two get replaced. The documentation states that all should be replaced.

fb55:
Unable to reproduce, closing.

PR #5431, opened 2026-08-13
fix(manipulation): copy the content when replacing several elements

maximilliangrand:
Refs #687.

`replaceWith` reuses the same replacement nodes for every selected element. Existing node content therefore moves from target to target, leaving only the last replacement. For example, replacing two `<p>` elements with one existing `<b>` should produce two `<b>` elements; previously only one remained. String content already works because it is parsed separately for each target.

This applies Cheerio's existing insertion convention: deep-clone node content for every target except the last, which receives the originals. Documentation and tests cover Cheerio objects, raw node arrays, descendants, and callback content.

The final selection index still owns the originals even when that target has been removed. The content is consumed and the callback runs for that detached target. Self-replacement now preserves the target's insertion position and removes it exactly once, preventing an adjacent sibling from being deleted. Regression tests check node identities, order, parents, and sibling links for first/middle/last self-replacement and content containing the target with its siblings.

Callback cloning remains consistent with Cheerio's other insertion methods. This differs from jQuery's per-callback insertion behavior; the PR does not claim complete jQuery parity for overlapping target/content selections.

Validation on Node 26.7.0:

- Previous PR head 15978163 with the new self-replacement regressions: 4 failed, 203 passed. Fixed manipulation suite: 207 passed.
- Full suite:15 files, 809 tests passed; no type errors.
- TypeScript, Biome, build, and ESLint excluding website pass.
- Built ESM self-replacement smoke checks and git diff --check pass.

Full repository ESLint requires the website's separate Astro dependencies, which were not installed in this checkout; its failure is `File 'astro/tsconfigs/strict' not found`. Upstream workflows on the previous head expired while awaiting approval, so those failures do not represent executed tests. Fresh upstream CI still needs maintainer approval.

AI-assisted implementation and independent AI review; validation results above were run against the code.

### DIFF
--- a/src/api/manipulation.spec.ts
+++ b/src/api/manipulation.spec.ts
@@ -1624,6 +1624,25 @@ describe('$(...)', () => {
       expect($fruits.children()[2]).toBe($plum[0]);
     });
 
+    it('(elem) : should consume the replacement when the last element was removed', () => {
+      const $children = $fruits.children();
+      const $plum = $('<li class="plum">Plum</li>');
+
+      $children.last().remove();
+      $children.replaceWith($plum);
+
+      /*
+       * jQuery replaces the remaining elements with copies and still consumes
+       * the original: it is detached from its old position even though its
+       * target no longer has a parent to insert it under.
+       */
+      expect($fruits.children()).toHaveLength(2);
+      expect($('.plum')).toHaveLength(2);
+      expect($fruits.children()[0]).not.toBe($plum[0]);
+      expect($fruits.children()[1]).not.toBe($plum[0]);
+      expect($plum[0].parent).toBe(null);
+    });
+
     it('(elem) : should NOP if removed', () => {
       const $pear = $('.pear');
       const $plum = $('<li class="plum">Plum</li>');
--- a/src/api/manipulation.spec.ts
+++ b/src/api/manipulation.spec.ts
@@ -1574,6 +1574,56 @@ describe('$(...)', () => {
       expect($fruits.children()[1]).toBe($('.apple')[0]);
     });
 
+    it('(elem) : should replace every selected element with a copy', () => {
+      const $plum = $('<li class="plum">Plum</li>');
+
+      $fruits.children().replaceWith($plum);
+
+      const children = $fruits.children();
+      expect(children).toHaveLength(3);
+      expect($('.plum')).toHaveLength(3);
+      // Only the last element receives the original node
+      expect(children[0]).not.toBe($plum[0]);
+      expect(children[1]).not.toBe($plum[0]);
+      expect(children[2]).toBe($plum[0]);
+    });
+
+    it('(elem) : should copy the descendants of the replacement', () => {
+      const $plum = $('<li class="plum">Plum <b>!</b></li>');
+
+      $fruits.children().replaceWith($plum);
+
+      expect($.html($fruits)).toBe(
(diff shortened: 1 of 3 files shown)


maximilliangrand:
The observation is real and is now fixed in 15978163 — but the fix differs from the suggestion, because I checked the suggestion against jQuery 4 first and jQuery does something else.

**What jQuery actually does when the last target is parentless** (selection captured, its last element `.remove()`d, then `replaceWith($existing)`):

```
jquery : body = <b id="m">M</b><b id="m">M</b>
jquery : original connected? false
```

The attached targets get **copies**, and the original is **consumed** — detached from its old position without being inserted anywhere. That's jQuery's fragment semantics: content is always pulled out of the document up front, and the no-clone slot belongs to the last element of the full selection even when that element can no longer be replaced.

Computing `lastIdx` from targets with a parent — the suggested fix — would instead hand the *original* to the last attached target, which jQuery doesn't do. It would also make the result depend on which elements happen to be attached, so the same call could return different node identities for the same document.

So the fix keeps `lastIdx` as the last index of the selection and instead moves the consume (`updateDOM(dom, null)`) in front of the parent check, capturing `parent`/`siblings` first so that replacing an element with itself keeps working (the existing `(self)` test caught exactly that ordering mistake in my first attempt). Result on the probe scenario:

```
cheerio: body = <b id="m">M</b><b id="m">M</b>
cheerio: original parent = null
```

byte-identical to jQuery. This also means the function form is now invoked for parentless targets, which likewise matches jQuery — I verified with an instrumented probe that jQuery 4 calls the function for indices `[0, 1, 2]` when the last element is parentless.

The regression test covers the mixed selection: two copies inserted, original consumed (`$plum[0].parent === null`). Suite is 804/804 with the new test.

One boundary note for the record: a node freshly created with `$('<h1>…</h1>')` is *not* parentless in cheerio — it hangs off its parse root — so it never took the early-return path; the only ways to reach it are nodes that were `.remove()`d or previously replaced. That's why the repro uses `.remove()`.

maximilliangrand:
Fixed self-replacement without deleting neighboring nodes. The target stays attached while the other replacement nodes are detached, so its insertion position remains valid; it is then removed once and reinserted with no additional deletion.

Added regressions for replacing the first, middle and last child with itself, and for content containing the target and its siblings. They check every node's identity, order, parent and sibling links. All four cases fail on the previous PR head. Also added explicit callback-index coverage for a removed final target while retaining the full-selection clone/consumption behavior discussed above.

Validation: all 809 tests pass, type checking, Biome, build, and ESLint outside website pass. Built ESM smoke checks pass. Full repository ESLint still requires the website's separate Astro dependencies, which are not installed in this checkout. AI-assisted implementation received an independent AI review and the test verification above.

maximilliangrand:
Clarified the API and website documentation: callback results follow the same copy-all-but-last policy, and a detached final target consumes the original replacement without inserting it. Renamed the three test titles to match the actual copy/original assertions.

Added a multi-target callback-self-return test that checks unchanged HTML, every sibling link, original-node detachment and identity flags `[false, false, true]`. This preserves the callback cloning convention already described in this PR rather than introducing a different self-return policy. The earlier originals are detached during their own replacements, not by the final iteration.

The manipulation suite passes all 208 tests; TypeScript, Biome and library ESLint pass. This follow-up changes documentation and tests only. The separate website dependency limitation from the previous validation still applies. AI-assisted implementation and verification.

PR #5434, opened 2026-08-14
fix(traversing): reset matchUntil index per starting element

rajanpanth:
## Summary

Resets the internal `matchUntil` index per starting element during traversal so each root starts from a clean state.

## Why

Without resetting this state, multi-root traversal can carry stale indices and return inconsistent selections.


rajanpanth:
Closing this duplicate: this exact commit was already merged in #5416. Sorry for the duplicate submission.

PR #5448, opened 2026-08-22
fix: make removeAttr case-insensitive for HTML documents

### LINKED ISSUE #5257: Bug: .removeAttr() should be case-insensitive for HTML documents

rajanpanth:
Fixes #5257.

htmlparser2 lowercases attribute names at parse time for HTML documents (`lowerCaseAttributeNames` defaults to `htmlMode`), so `elem.attribs` only ever holds lowercase keys, and `removeAttr('CLASS')` looked up a key that cannot exist:

```js
const $ = cheerio.load('<div class="test"></div>');
$('div').removeAttr('CLASS');
$('div').attr('class'); // was "test", the removal silently no-opped
```

HTML attribute names are ASCII case-insensitive, so `removeAttr` now lowercases the name outside `xmlMode`, matching what the parser stored. XML documents keep case-sensitive behavior, locked by a test where `ID` and `id` coexist and only the exact-case one is removed.

Verification: 129 attribute tests pass (2 added); reverting the source change alone fails the new HTML-case test; eslint clean; full unit suite 795 passing.

Scope note: `attr('CLASS')` and `prop()` share the same case-sensitivity for reads and writes. Left out deliberately since the getter/setter semantics are a wider surface; happy to follow up if you want them aligned too.

<a href="https://cubic.dev/pr/cheeriojs/cheerio/pull/5448?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/api/attributes.spec.ts
+++ b/src/api/attributes.spec.ts
@@ -790,6 +790,21 @@ describe('$(...)', () => {
       expect($fruits.attr('id')).toBeUndefined();
     });
 
+    it('(key) : should remove case-preserved SVG attributes', () => {
+      const $svg = load('<svg viewBox="0 0 10 10"></svg>');
+      expect($svg('svg').attr('viewBox')).toBe('0 0 10 10');
+      $svg('svg').removeAttr('viewBox');
+      expect($svg('svg').attr('viewBox')).toBeUndefined();
+    });
+
+    it('(KEY) : should remove attributes created with mixed case via attr()', () => {
+      const $el = $('.apple');
+      $el.attr('DATA-X', 'value');
+      $el.removeAttr('DATA-X');
+      expect($el.attr('DATA-X')).toBeUndefined();
+      expect($el.attr('data-x')).toBeUndefined();
+    });
+
     it('(KEY) : should stay case-sensitive in XML documents', () => {
       const $xml = load('<food ID="fruits" id="other"/>', { xml: true });
       const $food = $xml('food');
--- a/src/api/attributes.spec.ts
+++ b/src/api/attributes.spec.ts
@@ -790,19 +790,31 @@ describe('$(...)', () => {
       expect($fruits.attr('id')).toBeUndefined();
     });
 
-    it('(key) : should remove case-preserved SVG attributes', () => {
+    it('(key) : should keep SVG attribute names case-sensitive', () => {
       const $svg = load('<svg viewBox="0 0 10 10"></svg>');
+      expect($svg('svg').attr('viewBox')).toBe('0 0 10 10');
+      // SVG attribute names are case-sensitive, so a lowercase lookup
+      // must not remove the case-preserved attribute.
+      $svg('svg').removeAttr('viewbox');
       expect($svg('svg').attr('viewBox')).toBe('0 0 10 10');
       $svg('svg').removeAttr('viewBox');
       expect($svg('svg').attr('viewBox')).toBeUndefined();
     });
 
-    it('(KEY) : should remove attributes created with mixed case via attr()', () => {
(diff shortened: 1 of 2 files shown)


rajanpanth:
Both findings were right, thank you. Verified each against the stored attribs:

- parse5 (the default HTML parser here) preserves the adjusted case of foreign-content attributes, so `<svg viewBox>` stores the key as `viewBox`.
- `attr('DATA-X', 'value')` stores the caller's name verbatim.

Lowercasing only the lookup missed both. Pushed `a45bcf78`: removal in HTML mode now compares stored keys case-insensitively (`key.toLowerCase() === name.toLowerCase()`) instead of assuming a lowercase store, and XML mode stays exact-match. Two regression tests added: `removeAttr('viewBox')` on an SVG element, and `attr('DATA-X')` followed by `removeAttr('DATA-X')`. 131 attribute tests and the full 797-test unit suite pass; reverting the source alone fails the two new tests plus the original HTML-case test.

rajanpanth:
Second round addressed in the latest commit, all three were correct:

- **Foreign content stays case-sensitive.** Matching now checks `elem.namespace`: HTML-namespace elements compare ASCII case-insensitively, while SVG/MathML (and `xmlMode`) keep exact matching. `removeAttr('viewbox')` no longer deletes `viewBox`; `removeAttr('viewBox')` still does. This matches DOM semantics where only HTML-namespace elements lowercase attribute names.
- **ASCII-only folding.** Comparison uses an `asciiLowerCase` helper folding `A`-`Z` only, so `removeAttr('data-k')` no longer conflates a distinct `data-`+U+212A (Kelvin sign) attribute. Locked with a test using the literal character.
- **Missing attribs guard.** The scan bails when `elem.attribs` is absent, matching the old `removeAttribute` guard.

The two behavior tests now exercise cross-case both ways (stored `viewBox` vs lowercase lookup on SVG, stored `DATA-X` removed via `data-x` on HTML), addressing the test-strength note. 132 attribute tests and the 798-test unit suite pass; reverting the source alone fails the SVG and cross-case tests.

rajanpanth:
On the `cheerio/slim` point: I checked it directly rather than assuming, and the premise does not hold in the way the reports describe.

htmlparser2 lowercases every parsed attribute name in HTML mode, foreign content included. Loading `<svg viewBox="0 0 1 1">` through `cheerio/slim` stores the key as `viewbox`, not `viewBox`, and `namespace` is `undefined`:

```
keys: [ 'viewbox' ],  ns: undefined
```

So there is no preserved case for the slim path to protect, and the parse path cannot regress. What genuinely differs is an attribute set programmatically on an element that happens to be SVG, for example `$('svg').attr('viewBox', 'x')` followed by `removeAttr('viewbox')`: with parse5 that stays exact-match via `namespace`, with slim it now folds, because htmlparser2 provides no namespace information at all.

I chose to keep folding when `namespace` is absent, since in that mode the whole tree is being treated as HTML and slim users otherwise keep the original #5257 bug. Added a comment at the branch recording exactly that trade-off.

Happy to switch to exact matching whenever `namespace` is undefined instead, which is more conservative but leaves #5257 unfixed for `cheerio/slim` consumers. Maintainer call, and a one-line change either way.

132 attribute tests and the 798-test unit suite pass; biome and eslint clean.

PR #5449, opened 2026-08-22
fix: split style declarations at top-level semicolons only

### LINKED ISSUE #5410: .css() splits a declaration when a data URI's semicolon is followed by a colon
`.css()` still splits a declaration when the part after a semicolon contains a colon, and rewriting an unrelated property then corrupts the `style` attribute.

Tested on cheerio 1.2.0 (and current `main`).

### Reproduction

```js

const style = `background:url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg'/>")`;
const $ = cheerio.load(`<div style='${style}'></div>`);
const $div = $('div');

console.log($div.css());
$div.css('color', 'red');
console.log($div.attr('style'));
```

### Expected

```js
{ background: `url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg'/>")` }
// background: url("data:image/svg+xml;utf8,<svg xmlns='http://www.w3.org/2000/svg'/>"); color: red;
```

### Actual

```js
{
  background: 'url("data:image/svg+xml',
  "utf8,<svg xmlns='http": "//www.w3.org/2000/svg'/>\")"
}
// background: url("data:image/svg+xml; utf8,<svg xmlns='http: //www.w3.org/2000/svg'/>"); color: red;
```

Two separate problems:

1. The declaration is split at the `;` inside the data URI, producing a bogus second property.
2. Because `.css(prop, val)` re-serialises everything it parsed, setting an *unrelated* property writes the mangled value back to the element. The URL now contains `; ` and `: `, so the image silently stops loading. Nothing in the caller's code touched `background`.

### Why the #1134 fix doesn't cover this

#1134 (and #907 before it) fixed the same split for `url(data:image/png;base64,…)`. The fix keeps a fragment that has no colon as a continuation of the previous value, which works for `base64,…`. But a data URI whose text after the `;` *does* contain a colon takes the other branch and becomes a new property. `xmlns='http://www.w3.org/2000/svg'` is the common case here, since inline SVG data URIs almost always carry that attribute.

The same applies to any quoted value holding a `;` and a `:`:

```js
cheerio.load(`<li style="content: 'a;b:c'">`)('li').css();
// { content: "'a", b: "c'" }   expected: { content: "'a;b:c'" }
```

Per CSS syntax, a `;` or `:` inside a string or inside parentheses is part of the value and does not separate declarations. Checking the same inputs through a spec-compliant parser (postcss) returns a single declaration in each case.

### Fix

Skipping over quoted sections and parentheses when scanning for the separating `;` and `:` handles all of these, and subsumes the `base64` case that motivated #1134 rather than special-casing it. Happy to open a PR — I have the change and tests ready.

rajanpanth:
Fixes #5410.

`parse()` split the style attribute on every `;`, then healed fragments that contain no colon by appending them to the previous value. That healed `url(data:image/png;base64,...)`, but a data URI whose post-semicolon tail contains a colon, such as an SVG data URI with a namespace URL, was read as a fresh declaration:

```js
$div.css()
// { background: 'url("data:image/svg+xml', 'utf8,<svg xmlns=\'http': '//www.w3.org/2000/svg\'/>")' }
```

and any `.css('color', 'red')` write-back then serialized the corrupted map into the `style` attribute.

The fix splits declarations only at top-level semicolons: quotes (with backslash escapes) and parentheses are tracked, so `;` inside `url(...)` or inside a quoted string never terminates a declaration. The existing heal-append for genuinely malformed fragments (#1134) is untouched and its test still passes.

Two tests added: the issue's SVG data URI round-trips through `.css()` and survives an unrelated property write, and semicolons inside quoted strings (`content: 'a;b'`) stay in the value.

Verification: 18 css tests pass (2 added); reverting `css.ts` alone fails the new data-URI test; full unit suite 795 passing; eslint and prettier clean.

<a href="https://cubic.dev/pr/cheeriojs/cheerio/pull/5449?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/api/css.spec.ts
+++ b/src/api/css.spec.ts
@@ -178,6 +178,24 @@ describe('$(...)', () => {
         });
       });
 
+      it('should not treat a function whose name ends in url as url()', () => {
+        const el = cheerio('<div></div>');
+        el.attr('style', '--x: myurl(/* ( */ value); color: red');
+        expect(el.css()).toStrictEqual({
+          '--x': 'myurl(/* ( */ value)',
+          color: 'red',
+        });
+      });
+
+      it('should still recognize an uppercase URL() token', () => {
+        const el = cheerio('<div></div>');
+        el.attr('style', 'background: URL(https://host/*); color: red');
+        expect(el.css()).toStrictEqual({
+          background: 'URL(https://host/*)',
+          color: 'red',
+        });
+      });
+
       it('should treat comment markers inside url() as data', () => {
         const el = cheerio('<div></div>');
         el.attr('style', 'background: url(https://host/*); color: red');
--- a/src/api/css.spec.ts
+++ b/src/api/css.spec.ts
@@ -133,6 +133,26 @@ describe('$(...)', () => {
           'url(data:image/png;base64,iVBORw0KGgo)',
         );
       });
+
+      it('should not split data URIs whose tail contains a colon (#5410)', () => {
+        const background =
+          'url("data:image/svg+xml;utf8,<svg xmlns=\'http://www.w3.org/2000/svg\'/>")';
+        const el = cheerio('<div></div>');
+        el.attr('style', `background:${background}`);
+        expect(el.css()).toStrictEqual({ background });
+
+        el.css('color', 'red');
+        expect(el.css()).toStrictEqual({ background, color: 'red' });
+        expect(el.attr('style')).toStrictEqual(
+          `background: ${background}; color: red;`,
+        );
+      });
+
+      it('should not split on semicolons inside quoted strings', () => {
+        const el = cheerio('<div></div>');
+        el.attr('style', "content: 'a;b'; color: red");
+        expect(el.css()).toStrictEqual({ content: "'a;b'", color: 'red' });
+      });
     });
   });
 });
--- a/src/api/css.spec.ts
+++ b/src/api/css.spec.ts
@@ -153,6 +153,24 @@ describe('$(...)', () => {
         el.attr('style', "content: 'a;b'; color: red");
         expect(el.css()).toStrictEqual({ content: "'a;b'", color: 'red' });
       });
+
+      it('should not track parentheses inside comments', () => {
+        const el = cheerio('<div></div>');
+        el.attr('style', 'color: red/*(*/; background: blue');
+        expect(el.css()).toStrictEqual({
+          color: 'red/*(*/',
+          background: 'blue',
+        });
+      });
+
+      it('should treat escaped characters as data', () => {
+        const el = cheerio('<div></div>');
+        el.attr('style', 'font-family: foo\\(bar; color: red');
+        expect(el.css()).toStrictEqual({
+          'font-family': 'foo\\(bar',
+          color: 'red',
+        });
+      });
     });
   });
 });
--- a/src/api/css.spec.ts
+++ b/src/api/css.spec.ts
@@ -148,6 +148,36 @@ describe('$(...)', () => {
         );
       });
 
+      it('should recognize comments inside non-url functions', () => {
+        const el = cheerio('<div></div>');
+        el.attr('style', 'width: calc(1px + /* ( */ 2px); color: red');
+        expect(el.css()).toStrictEqual({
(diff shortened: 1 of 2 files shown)


rajanpanth:
Both review findings were valid, pushed `a05b5867` addressing them:

- **Comments**: `color: red/*(*/; background: blue` no longer hides the separator. The splitter now skips `/* ... */` wholesale, and since the quote branch is checked first, `/*` inside a quoted string stays literal.
- **Top-level escapes**: a backslash outside quotes now escapes the following character, so `font-family: foo\(bar; color: red` splits into two declarations instead of absorbing `color` into the font value.

Tests added for both shapes. 20 css tests pass; reverting the source alone fails both new tests plus the data-URI test; full unit suite 797 passing; eslint and prettier clean.

rajanpanth:
The P1 was valid, pushed a fix. `background: url(https://host/*); color: red` entered comment mode at the unmatched marker inside the unquoted url() token and absorbed the closing parenthesis plus everything after it. Per CSS syntax, comments are never recognized inside url() tokens, so comment recognition now applies at top level (depth 0) only. Inside parentheses that trade-off is safe for this splitter's purpose, since semicolons there are already ignored via the depth counter.

Test added for the url() shape. 21 css tests pass; reverting the source alone fails the new test; biome and eslint clean.

On the `String.raw` note: local `eslint` and `biome check` both pass on these files as written, so I left the fixtures as escaped literals.

rajanpanth:
Round three, both findings correct and now fixed in `c1ec4269`.

The root cause was mine: I used paren depth as a proxy for "inside `url()`", which suppressed comment recognition inside every function. Verified both regressions against the source before changing anything:

- `width: calc(1px + /* ( */ 2px); color: red` lost `color` entirely.
- `width: calc(100% /* it's 100 */); color: red` did too, via the apostrophe entering quote mode.

Now an explicit `urlDepth` tracks the unquoted `url(` token (quoted `url("...")` is a normal string and already handled by the quote branch), and comment suppression applies only there. Comments work normally inside `calc()`, `var()`, gradients, and everywhere else.

Also took the block-delimiter point: `{}` and `[]` open blocks too, so `--theme: { color: red; background: blue }; margin: 0` no longer splits inside the value.

Three tests added for those shapes, on top of the existing url()/comment/escape/data-URI ones. 24 css tests pass, full unit suite 801 passing, biome and eslint clean, and reverting the source alone fails 3 of the new tests.

rajanpanth:
The suffix-`url` finding was correct and is fixed in `d381b59c`. `isUrlToken` checked only the three characters before the parenthesis, so `myurl(` was classified as a URL token, comment handling was suppressed inside it, and the parenthesis in the comment unbalanced the depth counter:

```
--x: myurl(/* ( */ value); color: red    →  color was absorbed into --x
```

It now requires an identifier boundary before `url`, so `myurl(`, `noturl(` and friends take the normal function path. Case-insensitivity is preserved, locked with an `URL(` test.

I checked the other two suggestions and they do not reproduce on this branch, so I have not changed anything for them:

- `--value: noturl(/* ); */); color: red` already yields `--value: noturl(/* ); */)` and `color: red`.
- `url(https://h/a]b); color: red` already yields `background: url(https://h/a]b)` and `color: red`, and so does the sharper `url(https://h/a]b;c)` with a semicolon after the unmatched closer, for both `]` and `}`.

In those cases an unmatched closer does drop the depth early, but the declaration is then reassembled by the existing heal-append path for fragments without a colon, so the parsed result is correct. Happy to tighten the delimiter tracking anyway if you would rather not rely on that, though it would be a behavior-neutral refactor rather than a fix.

26 css tests pass (2 added), full unit suite 803 passing, biome and eslint clean, and reverting the source alone fails the new `myurl` test.

PR #5470, opened 2026-09-02
fix: derive data keys the way a browser derives dataset keys

rajanpanth:
### What

`.data()` reports keys that `.data(key)` cannot read back:

```js
const $el = cheerio.load('<div data-foo_bar="x"></div>')('div');
$el.data();          //=> { fooBar: 'x' }
$el.data('fooBar');  //=> undefined
```

`.data()` derives keys with `camelCase`, and `.data(key)` maps back with `cssCase`. `camelCase` treats `.`, `_` and a trailing separator as word boundaries, but `cssCase` only inserts a hyphen before a capital, so the mapping is not invertible: `fooBar` could have come from `foo-bar`, `foo_bar` or `foo.bar`.

### The reference

The HTML spec only treats **a hyphen followed by an ASCII lower-case letter** as a boundary, and jQuery does the same (`/-([a-z])/g`). I checked the current behaviour against `element.dataset` in Chrome rather than reasoning from the spec alone:

| attribute | `element.dataset` | cheerio before |
| --- | --- | --- |
| `data-foo_bar` | `foo_bar` | `fooBar` |
| `data-a.b` | `a.b` | `aB` |
| `data-foo-1` | `foo-1` | `foo1` |
| `data-foo-` | `foo-` | `foo` |
| `data-a-b-` | `aB-` | `aB` |
| `data-plain-key` | `plainKey` | `plainKey` |
| `data-foo--bar` | `foo-Bar` | `foo-Bar` |
| `data--foo` | `Foo` | `Foo` |
| `data-9x-y` | `9xY` | `9xY` |

Five of nine diverged. After the change all nine match, and every key `.data()` returns reads back through `.data(key)`.

### Scope

`camelCase` and `cssCase` are both `@private` and used only by the `data-*` path (`attributes.ts` lines 572 and 592), so nothing else changes.

The existing `camelCase` assertions in `src/utils.spec.ts` encoded the behaviour being fixed, so they are updated. I have kept them as explicit cases rather than deleting them, and added the browser-derived values.

### Verification

Full suite is 795 passing. I confirmed the three new tests fail with only `src/utils.ts` reverted, and that the other 129 tests in `attributes.spec.ts` pass either way, so they isolate the change. `biome check`, `eslint` and `tsc --noEmit` are clean.

<a href="https://cubic.dev/pr/cheeriojs/cheerio/pull/5470?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/api/attributes.spec.ts
+++ b/src/api/attributes.spec.ts
@@ -654,6 +654,28 @@ describe('$(...)', () => {
       }
     });
 
+    it('() : should store a derived __proto__ key as an own property', () => {
+      /*
+       * `data-__proto__` derives the literal key `__proto__`. Assigning it
+       * normally would hit the inherited setter and swap the data object's
+       * prototype instead of creating the key. A browser's `dataset` lists it
+       * as an own enumerable property with the raw value, verified in Chrome.
+       */
+      const $el = load(
+        `<div data-__proto__='{"admin":true}' data-x="1"></div>`,
+      )('div');
+      const data = $el.data();
+
+      expect(Object.keys(data)).toStrictEqual(['__proto__', 'x']);
+      expect(Object.hasOwn(data, '__proto__')).toBe(true);
+      expect(data).not.toHaveProperty('admin');
+      expect(({} as Record<string, unknown>)['admin']).toBeUndefined();
+
+      const fresh = load('<div data-__proto__="polluted"></div>')('div');
+
+      expect(fresh.data('__proto__')).toBe('polluted');
+    });
+
     it('() : every reported key should read back through .data(key)', () => {
       /*
        * The keys `.data()` hands back have to be usable with `.data(key)`, or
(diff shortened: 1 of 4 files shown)


rajanpanth:
Good catch on the `__proto__` path, addressed in `e210044`.

With keys now derived literally, `data-__proto__` does produce the exact key `__proto__`, and plain assignment reached the inherited setter instead of creating the property. Both cache assignment sites now write through `Object.defineProperty`, so the key becomes an own enumerable property holding the raw value.

I verified the reference behaviour in Chrome before choosing that shape: `element.dataset` on `<div data-__proto__='{"admin":true}'>` lists `__proto__` in `Object.keys`, `Object.hasOwn` is true, the value is the raw string, and `({}).admin` stays undefined. The new test asserts all four, plus the single-key `.data('__proto__')` read, and fails when only the source change is reverted.

PR #5471, opened 2026-09-02
fix: split the class attribute on ASCII whitespace only

rajanpanth:
### What

`rspace` is `/\s+/`, which is wider than the HTML spec's definition of a class boundary. `\s` also matches a vertical tab, a non-breaking space and the Unicode space separators, so cheerio sees two classes where a browser sees one:

```js
// U+00A0 between the two letters
cheerio.load('<p class="a b"></p>')('p').hasClass('a');  //=> true
```

A browser reports a single class literally named `a b`, so that is a false positive. The same `rspace` drives `splitNames`, so `addClass`, `removeClass` and `toggleClass` split on those characters too.

### The reference

I checked every candidate against `element.classList` in Chrome rather than reading the spec alone. Exactly five characters separate classes:

| separator | tab `U+0009` · line feed `U+000A` · form feed `U+000C` · carriage return `U+000D` · space `U+0020` |
| --- | --- |
| **not a separator** | `U+000B` vertical tab · `U+00A0` nbsp · `U+1680` · `U+2000` · `U+2028` · `U+2029` · `U+202F` · `U+205F` · `U+3000` · `U+FEFF` |

`\s` matches all fifteen, so ten of them were treated as boundaries incorrectly. After the change all fifteen agree with the browser.

### The change

One line, plus a comment explaining why the narrower set is correct:

```diff
-const rspace = /\s+/;
+const rspace = /[\t\n\f\r ]+/;
```

### Verification

Full suite is 796 passing. I confirmed the two new negative tests fail with only `src/api/attributes.ts` reverted, while the ASCII-whitespace test passes either way, so the new tests isolate the change rather than just restating existing behaviour. `biome check`, `eslint` and `tsc --noEmit` are clean.

<a href="https://cubic.dev/pr/cheeriojs/cheerio/pull/5471?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/src/api/attributes.spec.ts
+++ b/src/api/attributes.spec.ts
@@ -842,6 +842,45 @@ describe('$(...)', () => {
     });
   });
 
+  describe('class separators', () => {
+    /*
+     * The class attribute is split on ASCII whitespace only. Verified against
+     * `element.classList` in Chrome.
+     */
+    const asciiWhitespace = ['\t', '\n', '\f', '\r', ' '];
+    const notWhitespace = [
+      '\v',
+      '\u{A0}',
+      '\u{1680}',
+      '\u{2000}',
+      '\u{2028}',
+      '\u{2029}',
+      '\u{202F}',
+      '\u{205F}',
+      '\u{3000}',
+      '\u{FEFF}',
+    ];
+
+    it('should split on ASCII whitespace', () => {
+      for (const ch of asciiWhitespace) {
+        expect(load(`<p class="a${ch}b"></p>`)('p').hasClass('b')).toBe(true);
+      }
+    });
+
+    it('should not split on other whitespace', () => {
+      for (const ch of notWhitespace) {
+        expect(load(`<p class="a${ch}b"></p>`)('p').hasClass('b')).toBe(false);
+      }
+    });
+
+    it('should treat a non-breaking space as part of the class name', () => {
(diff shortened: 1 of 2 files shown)


rajanpanth:
Both review points addressed in `b5de52d`.

**splitNames trim**: it now trims with the same five ASCII whitespace characters the split uses, instead of `String.prototype.trim`. The reviewer's example was real: `toggleClass` splits its argument with `rspace` but the element's classes with `splitNames`, so on `class=" a"` the element side saw `a` while the argument side saw the full token, and the toggle added a duplicate instead of removing the class.

**Mutator coverage**: `addClass`, `removeClass` and `toggleClass` now have non-breaking-space token tests as asked.

One honest note on test strength: my first version of the edge test used `removeClass`, and reverting only the `splitNames` change left it green, because `removeClass` trims its argument through `splitNames` too, so both sides agreed by symmetry. I rewrote it around `toggleClass`, whose argument bypasses `splitNames`, and re-verified: reverting only the trim change now fails exactly that test, 131 others unaffected. Full suite 798 passing.

